PR #21, opened 2019-05-21
Fix compilation error due to `unsafe' keyword

unipro:
After revision f6c8909, the following compilation error occurs:
```
error: expected identifier, found keyword `unsafe`
 --> src/lib.rs:2:11
  |
2 | #![forbid(unsafe)]
  |           ^^^^^^ expected identifier, found keyword
```
I think you intended `[forbid (unsafe_code)] '.

### DIFF
--- a/src/connection_settings.rs
+++ b/src/connection_settings.rs
@@ -41,14 +41,14 @@ pub struct ConnectionSettings {
 
 impl ConnectionSettings {
     /// Timestamp in us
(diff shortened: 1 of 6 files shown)


russelltg:
Hmm, SRT uses 32-bit timestamps....

unipro:
Your reply is right, I know what you mean.
I changed the time used internally in the library to 64-bits timestamp.
My PR should not affect the transmission as the SRT Library still uses 32-bits timestamp when delivering data packets.

PR #25, opened 2020-01-09
POC: listen and connect as machine

robertream:
I spent a half day and redesigned/refactored the client/server connection setup to use an explicit state machine (i.e. mealy machine) instead of an implicit compiler generated state machine via async/await. This allowed for the removal of all dependencies on IO in the bulk of the listen and connect code, there’s just one external loop with a pattern match on the outputs of the state machine. The state machines take SRT packets and timer ticks in and outputs packets to send or errors to log. All inputs are also matched against current state using a few match statements. This approach to implementing protocols is frequently used in languages with advanced pattern matching facilities and algebraic data types (e.g. Erlang & typed functional languages). This makes inputs and outputs and transitions explicit. This would enable extremely fast thorough isolation testing (unit, property, generative, fuzz, etc.). I also find it much easier to reason about stateful network protocols this way. The only edge I ran up against is that I can't match against a borrow and a move in the same match statement in Rust, which required me to do more cloning and reference counting than I wanted to. If this is worth keeping around I'll come back and fix this later.


src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
+            remote,
+            local_sockid,
+            local_addr,
+            tsbpd_latency,
+        },
+        ConnectState::new(),
+    );
+
+    let mut tick_interval = interval(Duration::from_millis(100));
+    loop {
```
russelltg: I like this design--it's quite explicit. 

One thing that could be solved at the same time is right now it would be trivial to design a DOS attack on this by just sending an induction packet over and over, which would inhibit anone else from connecting. I think the resolution would be to have something like `HashMap<SocketAddr, ConnectState>`, and just use the address that resolves first.

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
 use crate::util::get_packet;
-use crate::{Connection, ConnectionSettings, SocketID, SrtVersion};
+use crate::{Connection, ConnectionSettings, DataPacket, SeqNumber, SocketID, SrtVersion};
 
-pub async fn connect<T>(
-    sock: &mut T,
+use ConnectError::*;
+use ConnectState::*;
+
+pub struct ConnectConfiguration {
```
russelltg: The choice of having this `pub` is interesting--it's not really usable from an outside module, as there is no way to construct it. Maybe a good way to do this would be to have a integration test that uses this state machine without the `async connect` function

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
+    UnsupportedProtocolVersion(u32),
+    // TODO: why don't we validate the cookie on responses
+    //#[error("Received invalid cookie handshake from [address], expected: {0} found {1}")]
+    //InvalidHandshakeCookie(i32, i32),
+    // TODO: why don't we validate we have an SRT response
+    //#[error("Expected SRT handshake request in conclusion handshake, found {0}")]
+    //SrtHandshakeExpected(HandshakeControlInfo),
+}
+
+pub struct Connect(ConnectConfiguration, ConnectState);
```
russelltg: Same for here as above with with respect to publicity. 

Also, I have a slight preference for using structs here instead of a tuple, i think `.field` is a bit easier than `.0`. But I'm willing to be convinced either way on this.

src/pending_connection/listen.rs
```diff
@@ -1,234 +1,269 @@
...
-                    ext_hs: Some(SrtControlPacket::HandshakeRequest(hs)),
-                    crypto_size,
-                    ..
-                } = shake.info
-                {
-                    (hs, crypto_size)
-                } else {
-                    bail!("Did not get SRT handshake request in conclusion handshake packet, using latency from this end");
+#[derive(Debug)]
+pub enum ListenError {
```
russelltg: Again, `non_exaustive` would be good

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
-                ext_config: None,
-            },
-            ..hs_info
-        }),
-    });
+}
 
-    sock.send((pack.clone(), remote)).await?;
+#[derive(Clone)]
+pub enum ConnectState {
```
russelltg: If we end up merging this I would like these tuple variants to have some docs, it's not immediately obvious what all of these are.

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
-                continue;
+        match result {
+            Ok(Some(packet)) => {
+                sock.send(packet).await?;
             }
-            if addr != remote {
-                warn!("Expected packet from {}, got {}", remote, addr);
-                continue;
+            Err(e) => {
+                eprintln!("{:?}", e);
```
russelltg: Don't like this, I would much prefer you use the log macros instead so it's configurable.

src/pending_connection/listen.rs
```diff
@@ -1,234 +1,269 @@
...
+    );
+
+    loop {
+        let packet = get_packet(sock).await?;
+        match listen.next_packet(packet) {
+            Ok(Some(packet)) => {
+                sock.send(packet).await?;
+            }
+            Err(e) => {
+                eprintln!("{:?}", e);
```
russelltg: Again, log macros

src/pending_connection/listen.rs
```diff
@@ -1,234 +1,269 @@
...
+            }
+            _ => {}
+        }
+        match listen.1.clone() {
```
russelltg: `if let`

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
             }
-            if info.info.version() != 5 {
-                warn!("Handshake was HSv4, expected HSv5");
-                continue;
+            _ => {}
+        }
+        match connect.1.clone() {
```
russelltg: `if let`

src/pending_connection/rendezvous.rs
```diff
@@ -1,167 +1,227 @@
...
+}
+
+impl Rendezvous {
+    pub fn new(config: RendezvousConfiguration) -> Self {
+        Self(config, Negotiating, rand::random())
+    }
+}
+
+#[derive(Debug)]
+pub enum RendezvousError {
```
russelltg: `non_exaustive`

src/pending_connection/connect.rs
robertream: I actually agree, when I have less than three values in a struct I'll prefer a tuple, but then, I'll switch to struct when I see patterns of direct field access instead of restructuring.

robertream: I actually agree, when I have less than three values in a struct I'll prefer a tuple, but then, I'll switch to struct when I see patterns of direct field access instead of de-structuring.

src/pending_connection/listen.rs
robertream: ok

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
...
-                    info @ HandshakeControlInfo {
+impl ConnectState {
+    pub fn new() -> ConnectState {
+        Configured(rand::random())
+    }
+}
+
+#[derive(Debug)]
+pub enum ConnectError {
+    //#[error("Expected Control packet, expected: {0} found: {1}")]
```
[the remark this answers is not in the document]
russelltg: It looks good for now--error handling could be a different PR.

src/pending_connection/connect.rs
russelltg: Yeah. out of the scope of this PR though.

russelltg:
Could you fix the clippy lints that you violated shown in azure?

The enum variant difference I'm not too worried about, feel free to suppress them. You need to eventally go though all the variants so it's not a big deal. 

The clone ones would be good to fix though.

russelltg:
Awesome work, thank you.

PR #26, opened 2020-01-15
POC: another example applying to sender

robertream:
I pulled out the core of UDT Sender's Algorithm and separated it from tokio, futures and streams and made it a separate module, this will not only make it easier to isolation test, it will also make it reusable with plain old system threads other async runtimes (e.g.) async-std, and it also allowed for simply mapping the letter of the spec to the code, so it's very readable. Look at SenderAlgorithmStep, SenderAlgorithmAction and Sender::next_algorithm_action, as well as the tokio::SrtCaller implementation.


src/protocol/mod.rs
```diff
@@ -0,0 +1,53 @@
+use crate::packet::{HandshakeControlInfo, ShakeType};
```
russelltg: I haven't been using `mod.rs` files in this project, preferring a `protocol.rs` in the parent directory. I don't really care either way but for consistency I would like it to be `protocol.rs`.

Same goes for `tokio/mod.rs`

tests/caller_listener.rs
```diff
@@ -0,0 +1,53 @@
...
+use srt::{ConnInitMethod, SrtSocketBuilder};
+use std::net::SocketAddr;
+use std::time::Duration;
+use tokio::time::delay_for;
+
+#[tokio::test]
+async fn caller() -> Result<(), Error> {
+    env_logger::init();
+
+    let _local = "0.0.0.0:0".parse::<SocketAddr>().unwrap();
```
russelltg: This line can be deleted

src/tokio/caller.rs
```diff
@@ -0,0 +1,135 @@
...
+use std::net::SocketAddr;
+use std::time::{Duration, Instant};
+use tokio::net::UdpSocket;
+use tokio::time::delay_until;
+use tokio_util::udp::UdpFramed;
+
+pub struct SrtCaller {
+    close_tx: Option<oneshot::Sender<()>>,
+    srt_send_tx: mpsc::Sender<(Bytes, Instant)>,
+    _task: tokio::task::JoinHandle<()>,
```
russelltg: Any reason that this is stored? Looking at the [docs](https://docs.rs/tokio/0.2.9/tokio/task/struct.JoinHandle.html) it seems like it detaches when it gets dropped, so if we're not joining on it dropping it seems fine.

src/tokio/caller.rs
```diff
@@ -0,0 +1,135 @@
...
+            _task: task,
+        })
+    }
+
+    pub async fn send(&mut self, data: Bytes) -> Result<(), mpsc::SendError> {
+        self.srt_send_tx.send((data, Instant::now())).await
+    }
+
+    pub async fn shutdown(&mut self) -> Result<(), Error> {
+        self.close_tx = None;
```
russelltg: A comment here saying this is droping the `close_rx` would be good, it's not that explicit

src/tokio/caller.rs
```diff
@@ -0,0 +1,135 @@
...
+use futures::select;
+use futures::{FutureExt, SinkExt, StreamExt};
+use log::debug;
+use std::net::SocketAddr;
+use std::time::{Duration, Instant};
+use tokio::net::UdpSocket;
+use tokio::time::delay_until;
+use tokio_util::udp::UdpFramed;
+
+pub struct SrtCaller {
```
russelltg: Not sure what this represents that the builder type doesn't.

Would the eventual goal to be move this into builder? Or maybe move this to `tokio::Builder`, to allow for a potential `async_std::Builder` as well?

src/lib.rs
```diff
@@ -21,6 +22,7 @@ mod socket_id;
 mod srt_congest_ctrl;
 mod srt_version;
 mod stats;
+pub mod tokio;
```
russelltg: With this new split, I would appreciate making tokio an optional dependency at some point. 

Probably something for a future PR though.

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+            buffer: Default::default(),
+            next_seq_number: settings.init_seq_num,
+            next_message_number: MsgNumber::new_truncate(0),
+        }
+    }
+
+    /// In the case of a message longer than the packet size,
+    /// It will be split into multiple packets
+    pub fn push_data(&mut self, data: (Bytes, Instant)) {
+        let (mut payload, time) = data;
```
russelltg: Put this in the parameter list: `(mut payload, time): (Bytes, Instant)`.

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+        let mut location = PacketLocation::FIRST | PacketLocation::LAST;
+        while !payload.is_empty() {
+            // if we need to break this packet up
+            if payload.len() > self.max_packet_size {
+                let slice = payload.slice(self.max_packet_size..payload.len());
+
+                self.begin_transmit(time, slice, location);
```
russelltg: The `LAST` flag will get set on the first one. I think the logic in this function needs some close thinking. Maybe a unit test would help.

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+                payload = payload.slice(0..self.max_packet_size)
+            } else {
+                location = location | PacketLocation::LAST;
+                self.begin_transmit(time, payload, location);
+                break;
+            }
+        }
+    }
+
+    fn begin_transmit(&mut self, time: Instant, payload: Bytes, location: PacketLocation) {
```
russelltg: A note about the precondition of `payload.len() < max packet size` would be good. Or an assertion

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+    }
+
+    pub fn release_acknowledged_packets(&mut self, acknowledged: SeqNumber) {
+        while acknowledged > self.first_seq {
+            self.buffer.pop_front();
+            self.first_seq += 1;
+        }
+    }
+
+    pub fn get<'a, I: Iterator<Item = SeqNumber> + 'a>(
```
russelltg: A comment describing what this does would be good. Something like iterator of indicies relative to the begin -> data packets

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+
+    pub fn push_back(&mut self, packet: DataPacket) {
+        self.list.push_back(packet);
+    }
+
+    pub fn pop_front(&mut self) -> Option<DataPacket> {
+        self.list.pop_front()
+    }
+
+    pub fn remoeve_acknowledged_packets(&mut self, acknowledged: SeqNumber) -> u32 {
```
russelltg: I think this is a typo

Also a comment saying what this returns would be good, it's not obvious from the signature.

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+        self.buffer.push_back(data);
+    }
+}
+
+pub struct LossList {
+    pub list: VecDeque<DataPacket>,
+}
+
+impl LossList {
+    pub fn new(_settings: &ConnectionSettings) -> Self {
```
russelltg: Any reason this takes in this?

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+        self.list.back()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.list.is_empty()
+    }
+}
+
+#[derive(Debug)]
+pub enum SenderError {}
```
russelltg: This needs fleshing out but you know that

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+            est_link_cap: 0,
+            lost_packets: 0,
+            retrans_packets: 0,
+            recvd_packets: 0,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum SenderAlgorithmStep {
```
russelltg: A comment with a link to the spec would be good here, so it can be deciphered what each step means. Alternatively, you could come up with descriptive names and leave the step numbers in the comments

Also, a comment with what the `Instants` mean might be good

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+    /// The buffer to store packets for retransmission, sorted chronologically
+    send_buffer: SendBuffer,
+
+    /// The buffer to store the next control packet
+    control_buffer: Option<ControlPacket>,
+
+    /// The buffer to store packets for transmission
+    transmit_buffer: TransmitBuffer,
+
+    // 1) Sender's Loss List: The sender's loss list is used to store the
```
russelltg: Again, good to have a link to the spec you copied this from (i'm assuming https://tools.ietf.org/html/draft-gg-udt-03)

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+            lr_acked_packet: settings.init_seq_num,
+            lr_acked_ack: -1, // TODO: why magic number?
+            congestion_control,
+            control_buffer: None,
+            transmit_buffer: TransmitBuffer::new(&settings),
+        }
+    }
+
+    pub fn next_algorithm_action<Now: Fn() -> Instant>(
+        &mut self,
```
russelltg: I would not have guessed this would take `&mut self`, maybe a different function name would be more appropriate.

src/protocol/sender.rs
```diff
@@ -0,0 +1,622 @@
...
+        }
+
+        Ok(None)
+    }
+
+    fn handle_data_packet(&mut self, _packet: DataPacket, _now: Instant) -> SenderResult {
+        Ok(None)
+    }
+
+    fn send_control(&mut self, control: ControlTypes, now: Instant) -> SenderResult {
```
russelltg: My understanding is the `SenderResult` returns a packet to be sent to the receiver--why does this function not return that packet?

src/tokio/caller.rs
```diff
@@ -0,0 +1,135 @@
...
+        use SenderAlgorithmAction::*;
+        loop {
+            let (current_step, next_action) = self.sender.next_algorithm_action(|| Instant::now());
+            debug!("{:?}: {:?}", current_step, next_action);
+            match next_action {
+                Continue => {}
+                SendControl((packet, to)) => self.send_control(packet, to).await,
+                SendData((packet, to)) => self.send_data(packet, to).await,
+                WaitUnitlAck => self.wait().await,
+                WaitUntilData => self.wait().await,
```
russelltg: Combine these cases

src/tokio/caller.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    async fn wait_until(&mut self, t: Instant) {
+        // TODO: get rid of the panics with legit error handling, e.g. an error channel on SrtCaller
+        let result = select! {
+            _ = &mut self.close_rx => self.sender.handle_close(Instant::now()),
+            _ = delay_until(t.into()).fuse() => self.sender.handle_timer(Instant::now()),
+            packet = crate::util::get_packet(&mut self.socket).fuse() => self.sender.handle_packet(packet.unwrap(), Instant::now()),
+            data = self.srt_send_rx.next().fuse() => self.sender.handle_send_rx(data.unwrap(), Instant::now()),
+        };
+        let _ = result.map_err(|error| eprintln!("{:?}", error));
```
russelltg: Prefer `log`

src/tokio/caller.rs
```diff
@@ -0,0 +1,135 @@
...
+use futures::channel::{mpsc, oneshot};
+use futures::future::Fuse;
+use futures::select;
+use futures::{FutureExt, SinkExt, StreamExt};
+use log::debug;
+use std::net::SocketAddr;
+use std::time::{Duration, Instant};
+use tokio::net::UdpSocket;
+use tokio::time::delay_until;
+use tokio_util::udp::UdpFramed;
```
russelltg: I would appreciate it if you cleaned these up.

src/lib.rs
```diff
@@ -12,6 +12,7 @@ mod msg_number;
 mod multiplex;
 mod packet;
 mod pending_connection;
+pub mod protocol;
```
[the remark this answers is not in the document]
robertream: 1) I wrote the code without too much attention to code cleanliness. My first priority was to provide a working refactored/redesigned proof of concept. I'm willing to concede whatever organization you wish, though the way I was thinking about it was to separate the packet/wire codec (i.e. the messages) from the protocol (i.e. algorithms, state machines, and supporting data structures), and one or more separate "API" top level libraries: srt-tokio, srt-async-std, srt-std, etc. that adapt the protocol to the runtime. No need for "features" to manage dependences. I really do dislike how tokio uses flags to manage what could otherwise be managed using basic dependencies.

2) The GitHub UI seems to be awkward here, I'm not sure how to directly comment on your initial code comment above, so I'll do so here: The theme of the design I had in mind was to invert control over non-deterministic operations (i.e. effects: querying system clock, network I/O, IPC, timers). The results of these non-deterministic operations also impact the state machine as new inputs, it's a feedback loop. The outputs of the state machine are not merely historical data about what happened inside the state machine (e.g. events), rather, they are delineated/explicit actions to be taken by an outer evaluation/execution context, and the result of the actions produce non-deterministic feedback, new input for the state machine. I believe coroutines or effect types were more what I was thinking of than events. I did consider using an "event" or "result" type/enum instead of the "handle_XX()" methods, but I didn't end up exploring that too far.

src/pending_connection/connect.rs
```diff
@@ -1,208 +1,261 @@
 use std::net::{IpAddr, SocketAddr};
```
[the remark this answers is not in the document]
robertream: yes

src/srt_congest_ctrl.rs
```diff
@@ -13,3 +13,10 @@ impl CongestCtrl for SrtCongestCtrl {
         10_0000
     }
 }
+
+impl SrtCongestCtrl {
+    #[allow(non_snake_case)]
+    pub fn SDN(&self) -> Duration {
```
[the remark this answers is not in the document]
robertream: I understand, send_interval and SDN were the same thing, it was an experiment to provoke discussion. How close can I make the code look like the spec. Excess or non-essential complexity is the distance between the concepts/model of the domain and the concepts and code in the implementation. Using as close to the same language in the code as is in the model can help us differentiate one from the other. In this case the spec and ref-implementation guide what the domain model is. The fact that it wasn't part of the trait was due to expediency. I would also suggest we avoid both ad-hoc and parametric polymorphism in the core domain model. Although I haven't gone too deep into the congestion control algorithms yet, combination of cheap sum types (enums) and pattern matching can actually provide some benefits over polymorphism. I replaced the ad-hoc polymorphism you were using in the mid session connection handshaking implementation with an enum. IMO it's more clear, explicit, because there are less layers of non-essential indirection.

src/protocol/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::ControlPacket;
+use std::net::SocketAddr;
+use std::time::Instant;
+
+pub mod sender;
+
+pub enum Handshake {
+    Caller,
```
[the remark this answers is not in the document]
robertream: Again I was experimenting with names a bit to provoke discussion. I also needed something to use as the top level API to drive a test. I hadn't settled on a name yet and went with caller because Caller-Listener is what the spec uses in this scenario. Upon further thought I actually think ultimately where I would suggest we go is to mirror the TCP implementation in tokio, it has TcpListener and TcpStream. So we'd have a SrtListener and SrtStream. SrtListener would be the only way to start a listener connection. It would have an ccept() and incoming() implementation that allow provide an SrtStream for each new connection, just like TcpListener, and SrtStream would have two connection methods: call() and rendezvous(). Would that work for you?

src/protocol/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+        &self,
+        from: SocketAddr,
+        handshake: HandshakeControlInfo,
+    ) -> Result<Option<(ControlPacket, SocketAddr)>, E> {
+        match (self, handshake.shake_type) {
+            (Handshake::Rendezvous(packet), ShakeType::Conclusion) => {
+                Ok(Some((packet.clone(), from)))
+            }
+            (Handshake::Listener(packet), _) => Ok(Some((packet.clone(), from))),
+            (Handshake::Caller, _) | (Handshake::Rendezvous(_), _) => Ok(None),
```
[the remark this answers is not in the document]
robertream: Fair, though it's a objective trade off between being explicit and succinct, I also may have left it there because Rendezvous actually should be doing something else, but I neglected to add a TODO comment.

src/protocol/sender.rs
robertream: Yes, good point. Though I was following the lead of the existing code, it wasn't consistently provided elsewhere. I'll be sure to add links where suitable from now on.

src/protocol/sender.rs
robertream: yes, I agree naming is important for facilitating API intuition, maybe "poll" would be better? But I was actually thinking of it as an iterator (it does return Option<T>...), so I chose "next"...

src/protocol/sender.rs
robertream: I think I was being indecisive on this design choice, if I wanted to actually have the "handle" code returning control packets or rather buffering them for the outer "next_algorithm_action " function to then output. I'm starting to lean towards the later. In the "connection setup" code from my first POC, this was intuitive and easy because only one control message was ever required per call to the "handle" functions, but when more than one or two are returned, then it would require I return a list or vector of packets, which is clumsy in rust as well as expensive with all the heap allocation involved, so if I buffer them like we are data packets, all is good. After digging into the receiver algorithm further, it will require buffering, a NAK and ACK event resulting from the same timeout can cause multiple packets to be sent, so I'm not convinced buffering is the best approach for all the things, to keep things consistent. Sorry for the confusion.

src/tokio/caller.rs
robertream: good point, as I was refactoring, I was kind of holding out that there would actually be a difference, but nope..

src/tokio/caller.rs
robertream: Yes, I think I did this eprintln stuff to quickly filter out all the other noise. I would not check the code in like this for sure. I don't really like libraries that swallow errors and rely on logging, especially bespoke logging. Chose any random C library as an example... :) but I still am not positive how I'd suggest errors be surfaced from the API later...

src/tokio/caller.rs
robertream: in what way clean them up? line break between tokio and std includes for example?

russelltg:
If you could rebase this, I'll give it another reivew pass.

PR #27, opened 2020-01-25
Receiver

robertream:
This is an almost drop in replacement of the existing Receiver. I replaced it with a tokio::ReceiverStream implementation. I tried doing the same with the Sender but so far the SenderStream implementation I have seems quite delicate with the packet loss tests, but the ReceiverStream has been solid, so I'd like to get it reviewed and merged first.


russelltg:
If you would rebase, I'll give it a review.

rream-isp:
Ok, I rebased. I also got SenderStream working last weekend as well so I included that here.

src/protocol/receiver/mod.rs
```diff
@@ -149,70 +171,121 @@ where
...
-    pub fn settings(&self) -> &ConnectionSettings {
-        &self.settings
+    pub fn handle_shutdown(&mut self) {
+        self.shutdown_flag = true;
     }
 
-    pub fn remote(&self) -> SocketAddr {
-        self.settings.remote
-    }
+    // handles an incomming a packet
```
russelltg: minor--fix spelling

src/protocol/receiver/mod.rs
```diff
@@ -43,11 +90,16 @@ struct AckHistoryEntry {
...
 
-pub struct Receiver<T> {
+pub struct Receiver {
     settings: ConnectionSettings,
 
-    // Function to return handshakes with
-    hs_returner: Option<HandshakeReturner>,
+    handshake: Handshake,
+
+    timers: crate::protocol::receiver::ReceiveTimers,
```
russelltg: Why the full path?

src/protocol/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    pub fn reset(&mut self, now: Instant) {
+        self.last = now;
+    }
+
+    pub fn set_period(&mut self, period: Duration) {
+        self.period = period;
+    }
+
+    pub fn poll_next(&mut self, now: Instant) -> Option<Instant> {
```
russelltg: I don't like the naming--`poll*` functions seem to generally functions that return a `Poll<T>`, and this isn't any sort of function. Maybe like `check_expired` or something.

src/protocol/receiver/mod.rs
```diff
@@ -149,70 +171,121 @@ where
...
+            Close
+        } else if let Some(Packet::Control(packet)) = self.pop_conotrol_packet() {
+            SendControl((packet, self.settings.remote))
+        } else {
+            TimeBoundedReceive(self.next_timer(now))
+        }
     }
 
-    fn on_ack_event(&mut self, cx: &mut Context) -> Result<(), Error> {
+    fn on_ack_event(&mut self, now: Instant) -> Result<(), Error> {
```
russelltg: This can only return `Ok`, just make it `() -> ()`.

src/protocol/receiver/mod.rs
```diff
@@ -603,146 +608,79 @@ where
...
-            self.buffer,
+            self.receive_buffer,
         );
 
         Ok(())
     }
 
     // send a NAK, and return the future
-    fn send_nak<I>(&mut self, cx: &mut Context, lost_seq_nums: I) -> Result<(), Error>
+    fn send_nak<I>(&mut self, now: Instant, lost_seq_nums: I) -> Result<(), Error>
```
russelltg: Can only return `Ok`, can you fix it to be `() -> ()`

src/protocol/receiver/mod.rs
```diff
@@ -326,46 +399,46 @@ where
...
-            timestamp: now,
+            timestamp: ts_now,
         });
-        self.send_to_remote(cx, ack)?;
 
         Ok(())
     }
 
-    fn on_nak_event(&mut self, cx: &mut Context) -> Result<(), Error> {
+    fn on_nak_event(&mut self, now: Instant) -> Result<(), Error> {
```
russelltg: Also infallable, should have no return value

src/protocol/receiver/mod.rs
```diff
@@ -603,146 +608,79 @@ where
...
-                    // end of stream, shutdown
-                    pin.shutdown_flag = true;
+    fn pop_conotrol_packet(&mut self) -> Option<Packet> {
+        self.control_packets.pop_front()
+    }
 
-                    continue;
-                }
-            };
+    fn on_exp_event(&mut self, _now: Instant) -> Result<(), Error> {
```
russelltg: I'd imagine the impl will be infallable, so have it return `()`

src/protocol/receiver/mod.rs
```diff
@@ -149,70 +171,121 @@ where
...
+        }
+        if self.timers.exp.poll_next(now).is_some() {
+            // TODO: what about EXP?
+            self.on_exp_event(now).unwrap();
+        }
+        if self.timers.timeout.poll_next(now).is_some() {
+            self.on_timeout();
+        }
+
+        //   2) Start time bounded UDP receiving. If no packet arrives, go to 1).
```
russelltg: Comment is in a bit of a misleading spot

src/protocol/sender/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+use crate::protocol::handshake::Handshake;
+use crate::protocol::sender::buffers::*;
+use crate::protocol::Timer;
+use crate::Packet::*;
+use crate::{
+    CCData, CongestCtrl, ConnectionSettings, ControlPacket, DataPacket, Packet, SeqNumber,
+    SrtCongestCtrl, Stats,
+};
+
+pub mod buffers;
```
russelltg: Pub? Are the buffers part of the public interface?

src/protocol/sender/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+
+#[derive(Debug)]
+pub enum SenderError {}
+
+pub type SenderResult = Result<(), SenderError>;
+
+pub struct SenderMetrics {
```
russelltg: This seems very similar to `Stats`, there should only be one. I prefer the name `SenderMetrics`, considering receiver doesn't currently expose `Stats`.

src/protocol/sender/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+            ControlTypes::Srt(srt_packet) => self.handle_srt_control_packet(srt_packet),
+            // The only purpose of keep-alive packet is to tell that the peer is still alive
+            // nothing needs to be done.
+            // TODO: is this actually true? check reference implementation
+            ControlTypes::KeepAlive => Ok(()),
+        }
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn handle_ack_packet(
```
russelltg: This is a bit gross. Could you break out the Ack struct enum variant into it's own struct, maybe `AckControlInfo`?

src/tokio/receiver.rs
```diff
@@ -0,0 +1,114 @@
+use std::future::Future;
```
russelltg: This is a good first pass. I might try to implement this with `stream::unwrap` at some point, might be cleaner. 

Don't worry about it for this PR though.

src/lib.rs
```diff
@@ -76,12 +75,12 @@ pub use crate::connection::{Connection, ConnectionSettings};
...
+pub use crate::tokio::receiver::ReceiverStream;
 // TODO: remove
-pub use crate::sender::Sender;
 pub use crate::seq_number::SeqNumber;
 pub use crate::socket::SrtSocket;
 pub use crate::socket_id::SocketID;
 pub use crate::srt_congest_ctrl::SrtCongestCtrl;
 pub use crate::srt_version::SrtVersion;
 pub use crate::stats::Stats;
+pub use crate::tokio::sender::SenderStream;
```
russelltg: Could you rename to `SenderSink`, as it's a sink?

src/pending_connection/connect.rs
```diff
@@ -71,8 +71,8 @@ pub enum ConnectError {
 }
 
 pub struct Connect {
-    config: ConnectConfiguration,
-    state: ConnectState,
+    pub config: ConnectConfiguration,
+    pub state: ConnectState,
```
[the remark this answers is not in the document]
robertream: I am leaving the fields on the *Configuration structs public. I believe I made these changes so that I could extract this code to the protocol module. It don't think it's essential, ultimately I think there is a better design that doesn't require access to the handshake state, perhaps as a return value.

src/protocol/handshake/mod.rs
```diff
@@ -0,0 +1,18 @@
+use crate::packet::{ControlTypes, HandshakeControlInfo, ShakeType};
```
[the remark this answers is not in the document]
robertream: I was going to extract the three handshake algorithm modules into the protocol module. This was just the first step.

src/protocol/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn timestamp_from(&self, instant: Instant) -> TimeStamp {
+        let elapsed = instant - self.0;
+        elapsed.as_micros() as i32 // TODO: handle overflow here
+    }
+
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn instant_from(&self, timestamp: TimeStamp) -> Instant {
+        self.0 + Duration::from_micros(timestamp as u64)
+    }
+}
```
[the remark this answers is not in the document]
robertream: I'm fine with this idea, especially if there is no other use of the socket_start_time, but this could be part of another pull request.

src/protocol/sender/mod.rs
robertream: No, they aren't part of the public interface. I believe this was a residual of the process I took to refactoring the code. I did it one small step at the time running all the tests each change, which included extracting the buffer classes and associated code as intermediate steps. I neglected to go back and clean up the public interface of the modules. I'll fix this, but I imagine there will be further improvement to be had besides this. Consider this a first pass at improving the code. I share your desire for a "clean" public interface, but it will evolve over time. As I mentioned before, my intention is to have three high level modules, each with an appropriate public interfaces: a packet (or wire?) module, a protocol (e.g. state machines, algorithms, timers, etc.), and a tokio module (e.g. builders, errors, SrtListener and SrtStream, etc. to mimic the tokio TCP design).

src/protocol/sender/mod.rs
robertream: I completely agree, yet this is something I am confident I will do to when I lay into the packet module a bit more. I think there are other improvements to be made here. Essentially I'm saying yes, and I will do this later. It will be ugly for now until it gets better, but rest assured I will not let this languish long, it's just non-essential to the goal of this current refactoring/pull request.

src/tokio/receiver.rs
robertream: I agree, I prefer higher level functions/combinators for stream signaling, creation and transformation, but a side effect free protocol extraction refactoring was my main goal here.

src/protocol/sender/mod.rs
russelltg: Sounds reasonable.

src/protocol/sender/mod.rs
russelltg: :+1:

PR #29, opened 2020-02-13
Implement receiver timeout

russelltg:
@robertream Does this look correct?

I think I followed the spec and the tests still pass.

### DIFF
--- a/src/protocol/receiver/mod.rs
+++ b/src/protocol/receiver/mod.rs
@@ -257,9 +257,8 @@ impl Receiver {
         if self.timers.nak.check_expired(now).is_some() {
             self.on_nak_event(now);
         }
-        if self.timers.exp.check_expired(now).is_some() && self.on_exp_event(now) {
-            self.shutdown_flag = true;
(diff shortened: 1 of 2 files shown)


src/protocol/receiver/mod.rs
```diff
@@ -258,12 +257,9 @@ impl Receiver {
...
         }
-        if self.timers.exp.check_expired(now).is_some() {
-            // TODO: what about EXP?
-            self.on_exp_event(now);
-        }
-        if self.timers.timeout.check_expired(now).is_some() {
-            self.on_timeout();
+        if self.timers.exp.check_expired(now).is_some() && self.on_exp_event(now) {
+            self.shutdown_flag = true;
+            info!("16 consequtive exp events, timeout");
```
robertream: This does look correct, according to the spec. Also, just make on_exp_event conditionally set the shutdown_flag. No need to have it return bool and complicate the semantics, none of the other “event” functions have a return value

russelltg: Good point. Will do.

PR #31, opened 2020-03-12
Timestamp drift recovery

robertream:
I took an initial pass at timestamp drift recovery. The clock only gets adjusted during stable network conditions. This is done by ensuring that drift deviation is within a hard coded tolerance (currently 5ms). This avoids abrupt and sometimes obtuse clock adjustments during periods of acute latency, e.g. due to network congestion. The reference implementation doesn't do this, but it should. I left a TODO to remind us to make the drift deviation tolerance configurable.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -20,7 +20,7 @@ impl TimeBase {
     #[allow(clippy::trivially_copy_pass_by_ref)]
     pub fn timestamp_from(&self, instant: Instant) -> TimeStamp {
         if self.0 > instant {
-            0 - (self.0 - instant).as_micros() as i32
+            -((self.0 - instant).as_micros() as i32)
         } else {
             (instant - self.0).as_micros() as i32
         }
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -7,27 +7,82 @@ pub mod sender;
 /// Timestamp in us
 pub type TimeStamp = i32;
 
+/// Duration in us, e.g. RTT
+pub type TimeSpan = i32;
+
 #[derive(Copy, Clone, Debug)]
 pub struct TimeBase(Instant);
 impl TimeBase {
-    pub fn new() -> Self {
-        Self(Instant::now())
-    }
-
-    pub fn from_raw(start_time: Instant) -> Self {
-        let _ = Self::new();
+    pub fn new(start_time: Instant) -> Self {
         Self(start_time)
     }
 
     #[allow(clippy::trivially_copy_pass_by_ref)]
     pub fn timestamp_from(&self, instant: Instant) -> TimeStamp {
-        let elapsed = instant - self.0;
-        elapsed.as_micros() as i32 // TODO: handle overflow here
+        if self.0 > instant {
+            0 - (self.0 - instant).as_micros() as i32
+        } else {
+            (instant - self.0).as_micros() as i32
+        }
     }
 
     #[allow(clippy::trivially_copy_pass_by_ref)]
     pub fn instant_from(&self, timestamp: TimeStamp) -> Instant {
-        self.0 + Duration::from_micros(timestamp as u64)
+        if timestamp < 0 {
+            self.0 - Duration::from_micros(timestamp.abs() as u64)
+        } else {
+            self.0 + Duration::from_micros(timestamp as u64)
+        }
+    }
+
+    pub fn adjust(&mut self, delta: TimeSpan) {
+        match delta {
+            delta if delta > 0 => {
+                self.0 += Duration::from_micros(delta as u64);
+            }
+            delta if delta < 0 => {
+                self.0 -= Duration::from_micros(delta.abs() as u64);
+            }
+            _ => {}
+        }
+    }
+}
+
+#[cfg(test)]
+mod timebase {
+    use super::*;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn timestamp_roundtrip(expected_ts: i32) {
+            let timebase = TimeBase::new(Instant::now());
+
+            let ts = timebase.timestamp_from(timebase.instant_from(expected_ts));
+            assert_eq!(ts, expected_ts);
+        }
+
+        #[test]
+        fn timestamp_from(expected_ts: i32, n in -10i64..10) {
+            let now = Instant::now();
+            let timebase = TimeBase::new(now);
+            let delta = ((std::i32::MAX as i64 + 1) * 2 * n as i64) + expected_ts as i64;
+            let instant = if delta > 0 { now + Duration::from_micros(delta as u64) } else { now - Duration::from_micros(delta.abs() as u64) };
+            let ts = timebase.timestamp_from(instant);
+            assert_eq!(ts, expected_ts);
+        }
+
+        #[test]
+        fn adjust(drift: i16) {
+            let now = Instant::now();
+            let mut timebase = TimeBase::new(now);
+
+            let original_ts = timebase.timestamp_from(now);
+            timebase.adjust(drift as TimeSpan);
+            let ts = timebase.timestamp_from(now);
+
+            assert_eq!(ts, original_ts - drift as TimeSpan);
+        }
     }
 }
 
--- a/src/protocol/receiver/time.rs
+++ b/src/protocol/receiver/time.rs
@@ -1,0 +1,111 @@
+use std::time::{Duration, Instant};
+
+use stats::OnlineStats;
+
+use crate::protocol::{TimeBase, TimeSpan, TimeStamp};
+
+pub(crate) struct SynchronizedRemoteClock {
+    tolerance: Duration,
+    time_base: TimeBase,
+    stats: Option<OnlineStats>,
+}
+
+impl SynchronizedRemoteClock {
+    const MAX_SAMPLES: usize = 1_000;
+    const DRIFT_TOLERANCE: Duration = Duration::from_millis(5);
+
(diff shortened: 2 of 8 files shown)


src/protocol/mod.rs
```diff
@@ -7,27 +7,82 @@ pub mod sender;
...
+    pub fn new(start_time: Instant) -> Self {
         Self(start_time)
     }
 
     #[allow(clippy::trivially_copy_pass_by_ref)]
     pub fn timestamp_from(&self, instant: Instant) -> TimeStamp {
-        let elapsed = instant - self.0;
-        elapsed.as_micros() as i32 // TODO: handle overflow here
+        if self.0 > instant {
+            0 - (self.0 - instant).as_micros() as i32
```
russelltg: Why `0 -` and not just prefix `-`?

src/protocol/mod.rs
```diff
@@ -7,27 +7,82 @@ pub mod sender;
...
+            delta if delta < 0 => {
+                self.0 -= Duration::from_micros(delta.abs() as u64);
+            }
+            _ => {}
+        }
+    }
+}
+
+#[cfg(test)]
+mod timebase {
```
russelltg: I've been calling the test module `test`

src/protocol/mod.rs
```diff
@@ -7,27 +7,82 @@ pub mod sender;
 /// Timestamp in us
 pub type TimeStamp = i32;
 
+/// Duration in us, e.g. RTT
+pub type TimeSpan = i32;
```
russelltg: I'm assuming this isn't a `Duration` because durations aren't signed?

src/protocol/mod.rs
```diff
@@ -7,27 +7,82 @@ pub mod sender;
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod timebase {
+    use super::*;
+    use proptest::prelude::*;
+
+    proptest! {
```
russelltg: Never seen this crate before! Looks pretty neat though. I'll have to look into it more.

src/protocol/receiver/time.rs
```diff
@@ -0,0 +1,111 @@
...
+impl SynchronizedRemoteClock {
+    const MAX_SAMPLES: usize = 1_000;
+    const DRIFT_TOLERANCE: Duration = Duration::from_millis(5);
+
+    pub fn new(now: Instant) -> Self {
+        Self {
+            // TODO: Drift deviation tolerance should be parameterized.
+            //       It wasn't in the reference implementation, but I added it because the reference
+            //       implementation is susceptible to invalid clock adjustments during periods of
+            //       acute network latency
```
russelltg: What is the tolerance in the reference impl?

src/protocol/receiver/time.rs
```diff
@@ -0,0 +1,111 @@
...
+                self.time_base.adjust(drift);
+            }
+            Some(stats) => {
+                stats.add(drift);
+
+                if stats.len() < Self::MAX_SAMPLES {
+                    return;
+                }
+
+                if stats.stddev() < self.tolerance.as_micros() as f64 {
```
russelltg: What if it never gets under this threshold?

src/protocol/receiver/time.rs
```diff
@@ -0,0 +1,111 @@
...
+        self.stats = Some(OnlineStats::new());
+    }
+
+    pub fn instant_from(&self, ts: TimeStamp) -> Instant {
+        self.time_base.instant_from(ts)
+    }
+}
+
+#[cfg(test)]
+mod synchronized_remote_clock {
```
russelltg: Again, about test module name

src/protocol/sender/buffers.rs
```diff
@@ -36,7 +36,7 @@ impl TransmitBuffer {
 
     /// In the case of a message longer than the packet size,
     /// It will be split into multiple packets
-    pub fn push_message(&mut self, data: (Instant, Bytes)) {
+    pub fn push_data(&mut self, data: (Instant, Bytes)) {
```
russelltg: I'd prefer this keeps the push_message name, it's a bit more descriptive I think.

src/protocol/receiver/time.rs
```diff
@@ -0,0 +1,111 @@
...
+            //       implementation is susceptible to invalid clock adjustments during periods of
+            //       acute network latency
+            tolerance: Self::DRIFT_TOLERANCE,
+            time_base: TimeBase::new(now),
+            stats: None,
+        }
+    }
+
+    pub fn synchronize(&mut self, now: Instant, ts: TimeStamp) {
+        let drift = self.time_base.timestamp_from(now) - ts;
```
russelltg: Isn't it incorrect that these should be the same, shouldn't now be `0.5 * rtt` after `ts`?

src/protocol/mod.rs
robertream: Yes, I was intending to turn this into a legit struct with all the right traits, a [Whole Value](http://fit.c2.com/wiki.cgi?WholeValue). Because Duration isn't signed, and Instant subtraction returns Durations, Instant is an unnecessarily cumbersome time API for SRTs use cases. As we evolve TimeStamp to become a first class Whole Value, we should build up a corresponding type (i.e. TimeSpan) to represent an SRT time range/duration/span so we can do time arithmetic in a clear, succinct, and reliable way.

src/protocol/receiver/time.rs
robertream: If it never gets under the threshold, then either the network between your hosts or their clocks are erratic, so a sufficiently stable clock sync will not be feasible. This threshold is what is considered sufficiently stable. If clock drift has high deviation, then you have much larger issues. I have no experience using the reference SRT implementation in networks with high congestion or latency variance, yet, so I'm not actually sure this is essential, but I do have experience working with clock sync algorithms in general, and I anticipate this will be an issue. I'm fine with removing this behavior for now.

src/protocol/sender/buffers.rs
robertream: Ok, I'm not sure why I changed this. I agree.

robertream:
The specification and/or documentation is misaligned with the reference implementation. This may have lead to your confusion. I was confused as well, but as I was actually pleasantly surprised how simple the essence of the reference implementation's actual algorithm was, and it was in line with my intuition of what it ought to be. It is using a perfectly valid approach to synchronize real time streaming clocks and is entirely suitable for the purpose of SRT's TSBPB feature.

The RTT approach may have some technical merit that I am unaware of, but it is more complicated than it needs to be. The reference implementation's approach (and mine) is built on an assumption that both systems have some reasonably stable view of time passing **and** that the latency of the communication channel between them is reasonably stable. In this approach, the receiver simply maintains a mapping between SRT timestamps and the receiver's system clock. The mapping is initialized using an arbitrary packet's timestamp and arrival time. For subsequent packets, an expected timestamp is calculated by mapping from the packets' arrival time back to a timestamp. The drift is then calculated by comparing the expected timestamp against the packet's actual timestamp. A periodic average of this drift is then calculated (the reference implementation used both a sample limit and a time limit to delimit the period). The mapping is then periodically adjusted based on this average.

The reference implementation also includes a bunch of extra complexity to be sure, and as far as I was able to tell the complexity is a result of some convoluted attempt at slewing the receiver's clock. Having carefully read through the logic of algorithm though, and absent any tests or data that demonstrate otherwise, I don't think it actually works. Slew is also unnecessary and actually undesirable for real time streaming clock sync. If timestamps jump, i.e. drift radically, slewing only lengthens the impact on realtime playback, so unlike many well known time synchronization algorithms that include slew (like NTP for example), for real time streaming, the magnitude of clock adjustments are immaterial, but the magnitude of the latency variance (i.e. stability of the network) is. So the better approach is to only make adjustments to the mapping during periods of network stability (STC from MPEG-TS is a good example of this). The two system clocks should realistically never drift more than a few milliseconds per hour, and with packetized real time streaming protocols the receiver should have plenty of measurements every second to work with in order to detect these periods. There is no need to know what RTT actually is.

src/protocol/mod.rs
robertream: Will this work if I locate unit test modules near their corresponding "units" (e.g. structs, enums, functions, etc.). Rust modules are a bit heavy weight, and I would prefer to have larger module files that contain many "units", and having a unit test module per "unit" like this allowed me to easily isolate and run all the tests for a given "unit" without having to create a separate module per "unit". The naming of these modules is a secondary concern to me.

src/protocol/mod.rs
robertream: Hmm, I'm not sure...

src/protocol/receiver/time.rs
robertream: It has no tolerance, it adjusts the time without regard to drift deviation.

src/protocol/mod.rs
russelltg: Yeah, if there's a good reason then I'm fine with it.

I'm willing to discuss, but I generally prefer smaller files. I also so far prefer just having the unit tests at the bottom so it doesn't clutter the code, but I'm pretty new to rust and my preferences are still evolving :).

src/protocol/mod.rs
russelltg: Sounds good.

Something we can implement later.

src/protocol/receiver/time.rs
russelltg: No it makes sense. Accounting for drift seems unreasonable if there is high latency variation, and just keeping with the original clock is likely better than trying to sync.

Issue #5, opened 2018-06-06
Test default CC implementation


russelltg:
I'm pretty sure  it's wrong, wayyy too slow.

Ways to test this: sending big files over lossy connections? Should compare to reference UDT?

Also could be possible it's not worth implementing and this repo should be pure SRT.

robertream:
@russelltg I have a pull request that implements Live CC, I don't think File is worth it. You should close this.

russelltg:
(will close when we merge the PR)

PR #32, opened 2020-03-19
clock sync + congestion control

robertream:
I've only implemented the Live Congestion Control algorithm. I derived it from the reference implementation. I question ever supporting File, it is not a compelling use case for SRT.


russelltg:
Ugh--I see this supersedes the other PR?

You should have closed the other one. Can we merge that one first, or are there time changes here that aren't in the other one.

robertream:
The changes I made in the second PR were additive. I can close this one if
you’d like?

robertream:
Ok, I say merge the first PR, I've added the few changes you requested to this PR, all except the test module name change. I think that requires further coding standards discussion, test names don't break functionality, so I'd rather get the new functionality integrated now and refactor tests and modules later to comply with an explicit coding standard.

russelltg:
Could you rebase?

russelltg:
Fixes #5

Issue #34, opened 2020-05-04
Threading


MarcAntoine-Arnaud:
Hello,

I'm starting using the library, what do you mean by `The reference implementation of SRT requires 3 threads per sender and 5 threads per receiver` ?

Can you provide simple examples in a folder `examples` ? it can be run using `cargo run --example sender` for a filename `sender.rs`
It can send fake data and implement also a demo example.

MarcAntoine-Arnaud:
If I change the data size to >8000 bytes the sender stop avec 113 packets.
Is it linked to the threading part ?

russelltg:
Heh, I wrote that threading thing a while ago...I remember doing some sort of benchmark where i spun up a lot of senders and receviers (I would hope in the same process, or else it wouldn't be apples to apples, although I cannot say that for sure) and saw how many OS threads, and that was the number I got. 

The main thing I was trying to show by that (and should likely be clarified) is that this implementation uses rust async tasks, which are sort of like greenthreads, that are much lighter than os threads so it in theory should have better performance with thousands of connections, not that that's a huge concern for SRT anyways.

Thanks for the examples PR, there are some small examples in doc comments but i should write more of them in general. The tests are a decent place to start as well.

> change data size to >8000 bytes

What do you mean? Could you give a code example?

MarcAntoine-Arnaud:
Yes `async` is just awesome !

In the example, [here](https://github.com/media-io/srt-rs/blob/add_examples/examples/sender.rs#L18) I have changed to 8000 and it not works as mentionned.

russelltg:
Oh wow that's surprising, considering I have a test that's pretty much exactly that (message_splitting) except it's 500k and not 8k. 

I'll debug it tonight, thanks a ton for the report.

russelltg:
I'm going to close this, I've commented on your PR. Feel free to repoen if you have more questions, or open a new issue.

PR #35, opened 2020-05-04
add Sender and Receiver examples

MarcAntoine-Arnaud:
issue #34

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ features = ["full"]
 
 [dev-dependencies]
 hex = "0.4"
+futures-util = "0.3"
 rand_distr = "0.2"
 env_logger = { version = "0.7", default-features = false }
 proptest = "0.9.5"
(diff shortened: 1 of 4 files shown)


russelltg:
If you fix the formatting, I'll merge this.

Don't worry about some of the tests failing, some of them are pretty flaky and have been difficult.

russelltg:
If you look [here](https://docs.rs/futures/0.3.5/futures/sink/trait.SinkExt.html#method.send), there's a hint to why it didn't work well. It requires multiple packets to send the larger, packets, so it was exaserbated, but flushing after each packet isn't a very efficient way to go, and it must have gotten behind and started dropping packets.

russelltg:
I don't have access to push to your branch PR branch, so I'll post my sender here:

```rust
use bytes::Bytes;
use futures::stream;
use futures::{SinkExt, StreamExt};
use tokio::time::delay_for;

use srt::SrtSocketBuilder;
use std::io::Error;
use std::time::{Duration, Instant};

#[tokio::main]
async fn main() -> Result<(), Error> {
    let mut srt_socket = SrtSocketBuilder::new_listen()
        .local_port(3333)
        .connect()
        .await?;

    let mut stream = stream::unfold(0 as i32, |count| async move {
        print!("\rSent {:?} packets", count);
        delay_for(Duration::from_millis(10)).await;
        return Some((Ok((Instant::now(), Bytes::from(vec![0; 8000]))), count + 1));
    })
    .boxed();

    srt_socket.send_all(&mut stream).await?;
    Ok(())
}
```

PR #43, opened 2020-06-09
Adding example of a streamer looping over a set of media

nipierre:
- Example showing how to do a never-ending stream looping over a user-defined set of media
- Can choose the chunk size

### DIFF
--- a/srt-protocol/Cargo.toml
+++ b/srt-protocol/Cargo.toml
@@ -18,8 +18,8 @@ rand = "0.7"
 aes-ctr = "0.4"
 aes-soft = "0.5"
 block-cipher = "0.8"
-pbkdf2 = { version = "0.4", default-features = false }
-hmac = "0.8"
+pbkdf2 = { version = "0.5", default-features = false }
+hmac = "0.9"
 sha-1 = "0.9"
 bitflags = "1"
 


examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+        )
+        .arg(
+            Arg::with_name("chunk-size")
+                .short("c")
+                .help("Chunk size")
+                .takes_value(true),
+        )
+        .get_matches();
+
+    let port = matches.value_of("port").unwrap_or("3333").to_owned();
```
MarcAntoine-Arnaud, line 38: use [default_value()](https://docs.rs/clap/2.33.1/clap/struct.Arg.html#method.default_value) instead

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+    print!("Value for port: {}", port);
+
+    let url = format!("127.0.0.1:{}", port);
+    let mut srt_socket = SrtSocketBuilder::new_connect(&url).connect().await.unwrap();
```
MarcAntoine-Arnaud, line 42: replace unwrap with `except("message here")`

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let chunk_size = matches
+        .value_of("chunk-size")
+        .unwrap_or("10")
```
MarcAntoine-Arnaud, line 46: same here, use `default_value`

MarcAntoine-Arnaud: change default value to 1316, 10 bits is very small ...

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+        .unwrap();
+
+    print!("Stream hooked");
+
+    let files: Vec<_> = matches.values_of("files").unwrap().collect();
+
+    print!("{:?}", files);
+
+    loop {
+        for file in files.iter() {
```
MarcAntoine-Arnaud, line 58:
```rust
for file in &files {
}
```

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+    let chunk_size = matches
+        .value_of("chunk-size")
+        .unwrap_or("10")
+        .to_owned()
+        .parse::<u64>()
+        .unwrap();
+
+    print!("Stream hooked");
+
+    let files: Vec<_> = matches.values_of("files").unwrap().collect();
```
MarcAntoine-Arnaud, line 53: remove `: Vec<_>`

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    print!("{:?}", files);
+
+    loop {
+        for file in files.iter() {
+            print!("Open file {}", file);
+            let mut f = File::open(file)?;
+            let metadata = f.metadata()?;
+            let mut buffer = vec![0; chunk_size as usize];
```
MarcAntoine-Arnaud, line 62: set `usize` at the creation of the variable

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let files: Vec<_> = matches.values_of("files").unwrap().collect();
+
+    print!("{:?}", files);
+
+    loop {
+        for file in files.iter() {
+            print!("Open file {}", file);
+            let mut f = File::open(file)?;
+            let metadata = f.metadata()?;
```
MarcAntoine-Arnaud, line 61: not needed

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+            let mut buffer = vec![0; chunk_size as usize];
+
+            for _ in 0..metadata.len() / chunk_size {
+                f.read_exact(&mut buffer)?;
```
MarcAntoine-Arnaud, line 65: use `read()` and get the number of bytes readed.
If it's [0](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read), it's EOF.

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+                srt_socket
+                    .send((Instant::now(), Bytes::copy_from_slice(&buffer)))
```
MarcAntoine-Arnaud, line 67: use `Bytes::from(&buffer)` no copy needed.

examples/streamer.rs
```diff
@@ -0,0 +1,73 @@
...
+                    .await
+                    .unwrap();
```
MarcAntoine-Arnaud, line 69: use `?` to throw the error

russelltg:
Wow---thanks both of you!

This does have the same problem that @MarcAntoine-Arnaud's examples had--it uses the [`send`](https://docs.rs/futures/0.3.5/futures/sink/trait.SinkExt.html#method.send) function on `Sink`, which has the disclaimer in the docs:

> Note that, because of the flushing requirement, it is usually better to batch together items to send via send_all, rather than flushing between each item.

I'd probably use `futures::stream::unfold` to create a stream.

examples/streamer.rs
```diff
@@ -0,0 +1,75 @@
...
+            let mut f = File::open(file)?;
+
+            loop {
+                let mut buffer = vec![0; chunk_size];
+                let size = f.read(&mut buffer)?;
+                if size == 0 {
+                    break;
+                }
+                srt_socket
+                    .send((Instant::now(), Bytes::from(buffer)))
```
russelltg, line 70: Use send_all with a stream

russelltg:
I just did some restructuring, this should be moved into srt-tokio/examples

PR #55, opened 2020-07-21
Fix connection setup failure for security reasons

ian-spoonradio:
The client of the reference implementation cannot be connected to your implementation.
```
$ test-c-client 127.0.0.1 3333
srt startup
srt socket
srt remote address
srt setsockflag
srt connect
srt_connect: Connection setup failure: abort for security reasons
```

I think that the listener should use peer's ISN and send it back for security check.

### DIFF
--- a/srt-protocol/src/pending_connection.rs
+++ b/srt-protocol/src/pending_connection.rs
@@ -33,7 +33,6 @@ pub enum ConnectError {
 
 #[derive(Debug, Clone)]
 pub struct ConnInitSettings {
-    pub starting_send_seqnum: SeqNumber,
     pub local_sockid: SocketID,
     pub crypto: Option<CryptoOptions>,
     pub send_latency: Duration,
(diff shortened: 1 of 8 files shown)


russelltg:
Can you push an empty commit or something to re-run CI? I can't seem to re-run from pipelines.

russelltg:
Great observation here--something's still not quite right, the crypto exchange tests don't seem to work because now each side has a different idea of what the initial sequence number should be. 

I have tried to abstract the idea of HSv5 responders and initiators in hsv5.rs, but it seems like it's not *totally* the same for rendezvous and connect/listen--it seems like in rendezvous, each direction has their own initial sequence number, whereas listen and connect share the same initial sequence number for each direction.

russelltg:
I don't have access to your branch, so I can't put my changes here in this PR. I put my additions that have the tests passing and such in `fix_init_seq_number` branch. You can either merge that into your branch, or if it seems alright i'll just merge it in (which is what i'll probably do if you don't get back to me in a few weeks).

ian-spoonradio:
I don't understand enough of your code.
I tested the desired behavior in the `fix_init_seq_number` branch.
I expect you to fix it properly.

russelltg:
Just to confirm, that branch does work for you?

ian-spoonradio:
I did a test after merging that branch.
Native SRT app is connected well.
Thank you.

PR #59, opened 2020-08-10
fix init seq number

russelltg:
- Fix connection setup failure for security reasons
- Update PR to re-run CI
- Update PR to re-run CI
- Fix initial sequenc number problems

Based on #55 by @ian-spoonradio


russelltg:
This is fixed by #107

PR #60, opened 2020-08-24
implemented receiver peer idle timeout

robertream:
@russelltg I've been busy with many other things for many months now, but I had some time and energy today to sit down and code for fun, and this is what I came up with.

I saw that you removed EXP from ReceiveTimers. With this PR, I restored it. I implemented the peer idle timeout algorithm from the reference implementation too. I saw that you started implementing a Connection state machine. I think this is probably a good path to pursue, yet the EXP timer needs to be updated based on RTT, and RTT is calculated in the receive, so I feel this was the easiest spot to implement EXP based peer idle timeout. I'm open to extracting this eventually as Connection solidifies.

The PR also includes improvements to TimeSpan, TimeStamp, and TimeBase as well, including better test coverage. So it is worth merging if simply for this.


russelltg:
This is probably mostly good, it's a bit difficult to review because you moved the timespan implementation across files. 

EXP wise, you're right, but it needs to be in sender as well then....not exactly sure where it should be. Maybe it should be manually fed into the connection state machine, or maybe the connection state machine should receive the ACK packets as well. Not sure there

russelltg:
Hey, sorry this got left here neglected--I was working on some other timing stuff around the same time so was worried about conflicts then it slid under my radar for a while.

It's also difficult to review PRs with so many changes, but if there's still good stuff here I'd be willing to look at an updated PR.

robertream:
II was planning on updating this once the other PR is merged.

robertream:
Ok, so I think the Connection state machine needs to go away. My reasoning is as follows:
1. I propose that we only support bidirectional streams. 
2. Bidirectional streams will provide regular bidirectional Sender heart beat packets.
3. We only need one command packet per EXP interval to get through in order to keep a Receiver alive.
4. Even an input starved Sender should provide plenty of command packets per EXP interval.
5. Therefore we don't need the Sender to participate in the peer idle timeout algorithm.
6. The connection state can be derived from both Receiver and Sender state: e.g. one of them dies then the connection dies.

The benefits to the design as I see them are:
**Simplicity**
The current tokio code got quite complicated with the addition of the Connection algorithm, to implement the peer idle timeout algorithm correctly, it will have to get even more complicated. I think this is unnecessary complexity.
**Concurrency**
This design might allow for concurrent, low coordination, and lock free execution of both the Sender and Receiver algorithms (maybe, possibly, not 100% sure though). Of course, the reference implementation is not implemented this way, but it could, and I assert that this is also one factor that is contributing to all the lock and thread complexity in the codebase, which speaks to my prior point regarding simplicity. If we can avoid coordination and get equivalent (but not exactly the same) systemic behavior, we should do so.

A corollary is that at some point last year I was also looking into partitioning the packet data structures so that the Sender and Receiver specific packets were actually delineated in the types themselves.

<SoapBox>I've picked up a few design principles, or "rules of thumb", over the years of experience with data modeling in typed languages (...decades of experience actually, damn I'm old.. ha!) In languages that provide algebraic types, whenever there are match expressions on enums (i.e. sum types) that create partial functions (e.g. exceptions/panics or "this shouldn't happen" dead code comments/logging), this is a code smell, especially if the data types are internal to your own code. It is a glaring sign that your types might be better designed. They probably ought to be decoupled to better accommodate the intended needs of the usually singular point in your code base that consumes the sum types' data branches. Both the Receiver and Sender suffer from this code smell, therefore we ought to consider a refactor/redesign.

russelltg:
Only supporting bidirectional streams is fine by me, seems it would simplify things considerably and also make the tokio layer much simpler, as you note. 

> Even an input starved Sender should provide plenty of command packets per EXP interval.

I don't think this is the case--there's nothing for it to send! Senders do need to send KeepAlive's I'm pretty sure--but if we only support bidirectional sockets. Either senders or receivers can send KeepAlive's--for example the case of no data being sent (see the keepalive socket test). 

So my understanding is for fully correct KeepAlive behavior, we need to know the last total packet for anyone, not just receiver or sender. While full correctness isn't the most important, it seems like it'll send a KeepAlive crazy often....looked at wireshark to confirm:

image

> Therefore we don't need the Sender to participate in the peer idle timeout algorithm.

Why not? Say you have a unidirectional stream a -> b (so a's receiver state machine sees no packet), and b dies...it would be the sender that would have to deal with the timeout.

> The connection state can be derived from both Receiver and Sender state: e.g. one of them dies then the connection dies.

This is more or less true--the actual "this socket is fully closed" logic is much more intricate than I once thought, as you can see in the socket--would be great if we could clean it up though--and it shoudln't be in socket anyways. 

> This design might allow for concurrent, low coordination, and lock free execution of both the Sender and Receiver algorithms (maybe, possibly, not 100% sure though).

You mean without any queues? 

You may be right...I'm not totally sure. Something we eventually have to decide is how we can bring multiplexed connections more as a first class thing....I think we should mirror TCP servers for connections initialized as a listener (ie make all listeners multiplexed)--which would require queues as I assume we still want to spawn a tokio task for each connection (we may be able to put them all on the same task but it would scale a bit less well, not spreading over multiple cores. But it could be the state machine is so cheap that this overhead isn't really worth it?).

> If we can avoid coordination and get equivalent (but not exactly the same) systemic behavior, we should do so.

Agreed

> `<SoapBoxRant />`

You talking about Packet specifically? That has bothered me, but I'm not really sure how to handle it best....

srt-protocol/src/protocol/receiver/time.rs
```diff
@@ -189,42 +189,107 @@ impl Rtt {
...
+    pub fn check_ack(&mut self, now: Instant) -> Option<Instant> {
+        self.ack.check_expired(now)
+    }
+
+    pub fn check_nak(&mut self, now: Instant) -> Option<Instant> {
+        self.nak.check_expired(now)
+    }
+
+    pub fn check_peer_idle_timeout(&mut self, now: Instant) -> Option<Instant> {
+        self.exp.check_expired(now).and_then(|_| {
```
russelltg, line 256: I think `if ___.is_some() { }` is a bit clearer

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -429,6 +425,12 @@ impl Sender {
         }
     }
 
+    fn on_snd_event(&mut self, now: Instant) {
+        self.snd_timer.reset(now);
+        self.send_control(KeepAlive, now);
```
russelltg, line 430: Sending this every SND event is a LOT

russelltg (changes requested):

robertream:
> I don't think this is the case--there's nothing for it to send! Senders do need to send KeepAlive's I'm pretty sure--but if we only support bidirectional sockets. Either senders or receivers can send KeepAlive's--for example the case of no data being sent (see the keepalive socket test).

Here's a design heuristic for a keep alive though: If the keep alive is not frequent enough to actually keep a connection alive, what purpose does it serve? It could be removed and provide equivalent system behavior. If a 1s keep alive timer leads to frequent unexpected timeouts, then a better keep alive frequency out to be selected. One based on empirical data and perhaps an adaptive theoretical model or formula, otherwise a keep alive will not do what you think it does, and is therefore design, implementation and bandwidth waste.

I also don't disagree that SND was too frequent. I had neglected to copy the keepalive timer. The changes were hastilly thrown together.

> > Therefore we don't need the Sender to participate in the peer idle timeout algorithm.
> 
> Why not? Say you have a unidirectional stream a -> b (so a's receiver state machine sees no packet), and b dies...it would be the sender that would have to deal with the timeout.

I meant the EXP driven idle timeout algorithm. If the Sender is at least sending keep alive packets, then the Receiver of the  client on the other end of the connection can handle the idle timeout algorithm without coordination with its corresponding Sender.

> > The connection state can be derived from both Receiver and Sender state: e.g. one of them dies then the connection dies.
> 
> This is more or less true--the actual "this socket is fully closed" logic is much more intricate than I once thought, as you can see in the socket--would be great if we could clean it up though--and it shoudln't be in socket anyways.

I agree, it shouldn't be in socket. I have some thoughts on how to deal with this further, but that's a discussion for another issue.

> > This design might allow for concurrent, low coordination, and lock free execution of both the Sender and Receiver algorithms (maybe, possibly, not 100% sure though).
> 
> You mean without any queues?

No, I just mean no coordination between the Sender and Receiver of the same client.

> You may be right...I'm not totally sure. Something we eventually have to decide is how we can bring multiplexed connections more as a first class thing....I think we should mirror TCP servers for connections initialized as a listener (ie make all listeners multiplexed)--which would require queues as I assume we still want to spawn a tokio task for each connection (we may be able to put them all on the same task but it would scale a bit less well, not spreading over multiple cores. But it could be the state machine is so cheap that this overhead isn't really worth it?).

I agree, and I feel we ought to mirror the async/tokio TcpListener/TcpStream design closely. I don't have clear design ideas beyond that, except that by aiming to accommodate maximum concurrency, particularly between senders and receivers of the same client, we will avoid having non-essential design constraints holding us back from optimizing for performance.

> > `<SoapBoxRant />`
> 
> You talking about Packet specifically? That has bothered me, but I'm not really sure how to handle it best....

Create separate enums for Sender and Receiver inputs that eliminate most of the partial functions and dead branches in the match expressions. If there is an intersection between the two then consider if there is some missing, shared concept that could be extracted (i.e. a "connection")

russelltg:
Looking better. Can we reset the keepalive timer when we send a packet?

We would still unnecessarily send keepalive's on the receiving end, but that's a very low number of packets, nothing necessarily to worry about rn.

russelltg:
> No, I just mean no coordination between the Sender and Receiver of the same client.

I see, and agreed. Although I don't think this is entirely possible, because they need a sync'd RTT, which either could produce and both need. Could be okay if they maintain their own RTT though, I'm not really sure.

robertream:
> I see, and agreed. Although I don't think this is entirely possible, because they need a sync'd RTT, which either could produce and both need. Could be okay if they maintain their own RTT though, I'm not really sure.

Why does the Sender need RTT?

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -427,10 +432,14 @@ impl Sender {
 
     fn on_snd_event(&mut self, now: Instant) {
         self.snd_timer.reset(now);
-        self.send_control(KeepAlive, now);
         self.step = SenderAlgorithmStep::Step1;
     }
 
+    fn on_keepalive_event(&mut self, now: Instant) {
+        self.keepalive_timer.reset(now);
```
robertream, line 439: @russelltg the keepalive timer is reset here

russelltg: Right but we should also reset it when we send a packet

russelltg:
I think we should reset when we send a data packet as well--anything we send is an indication of the connection being alive.

russelltg:
Also you're right--sender does not RTT.

robertream:
I believe this was a spurious failure. I am uncertain how to retry the failed check without pushing a new commit.

PR #66, opened 2021-01-10
Fix parsing & encoding of stream id and expose it

### LINKED ISSUE #65: Exposing streamid
Would it be possible to expose the `streamid` described [here](https://github.com/Haivision/srt/blob/master/docs/AccessControl.md)? I see that it's already getting parsed in the handshake, but I'm not really sure how the best way to expose it would be. AFAICT it only arrives after a listener socket has received a connection.

fkaa:
The parsing and serialization of the stream id was not converting it to and from big-endian 4-byte-words.

It was also not properly set on the handshake response for a `Connection`.

Related #65

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
-[target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "linker=clang", "-C", "link-arg=-fuse-ld=lld"]
+#[target.x86_64-unknown-linux-gnu]
+#rustflags = ["-C", "linker=clang", "-C", "link-arg=-fuse-ld=lld"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
-#[target.x86_64-unknown-linux-gnu]
-#rustflags = ["-C", "linker=clang", "-C", "link-arg=-fuse-ld=lld"]
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-C", "linker=clang", "-C", "link-arg=-fuse-ld=lld"]
--- a/srt-protocol/src/packet/control/srt.rs
+++ b/srt-protocol/src/packet/control/srt.rs
@@ -210,24 +210,23 @@ impl SrtControlPacket {
             5 => {
                 // the stream id string is stored as 32-bit little endian words
                 // https://tools.ietf.org/html/draft-sharabayko-mops-srt-01#section-3.2.1.3
-                let mut bytes = Vec::with_capacity(buf.remaining());
-                let mut chunks = buf.chunk().chunks(4).peekable();
-
-                while let Some(&[a, b, c, d]) = chunks.next() {
-                    // make sure to skip padding bytes if any for the last word
-                    if chunks.peek().is_none() {
-                        match (d, c, b, a) {
-                            (_, 0, 0, 0) => bytes.push(d),
-                            (_, _, 0, 0) => bytes.extend(&[d, c]),
-                            (_, _, _, 0) => bytes.extend(&[d, c, b]),
-                            _ => {}
-                        }
-                    } else {
-                        bytes.extend(&[d, c, b, a]);
-                    }
+                if buf.remaining() % 4 != 0 {
+                    return Err(PacketParseError::NotEnoughData);
                 }
 
-                buf.advance(buf.remaining());
+                let mut bytes = Vec::with_capacity(buf.remaining());
+
+                while buf.remaining() > 4 {
+                    bytes.extend(&buf.get_u32_le().to_be_bytes());
+                }
+
+                // make sure to skip padding bytes if any for the last word
+                match buf.get_u32_le().to_be_bytes() {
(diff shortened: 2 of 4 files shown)


russelltg:
Had no idea the endianness was swapped, although I think that seems familiar.
How'd you know? If there's a spot in the srt source, could you add a permalink as a comment? Or even better, if you have a packet from wireshark from the original implementation or something that you could add as a unit test. 

I do see that the tests don't pass anymore, seems like it won't parse the packet that you're serializing. I'm sure it's something trivial, if you can't figure it out let me know and I'll be happy to take a look.

This compliments what i'm working on right now, which is a more complete framework for dealing with these streamid packets, in #58. It's getting close, and I'll publish some examples on how to do it and I'll link your issue to that PR as that's the one that properly exposes it (these packet structures you're editing are considered "internal" apis for SRT--someone using the API shouldn't have to deal with individual SRT packets.

russelltg:
I assume you're exposing connection through srt socket to be able to get to handshake and eventually the streamid? 

That's not really what I want, because that handshake will only have a packet in it if it's a listener, not a connector (that structure is for the SRT socket to know how to respond to incoming handshakes, in case their last handshake was dropped). I'd like it if you roll that back and wait until I finish the "real" solution.

fkaa:
I got it confused a bit, if it were big-endian it would not change the order of the character bytes. For whatever reason the stream id is sent as 4 byte little endian words (see [here](https://haivision.github.io/srt-rfc/draft-sharabayko-mops-srt.html#section-3.2.1.3)).

I saw it on the rust side when I was trying to stream SRT using OBS (which uses ffmpeg which uses libsrt) to a listener with srt-rs, and confirmed that it was being sent from ffmpeg with wireshark. If you still need a wireshark packet I can include it.

I only want the stream id yes, so I'll revert the API changes and wait for your upcoming changes instead.

russelltg:
That link is helpful, thank you. I think the test is failing because the padding gets placed in the wrong place for the last word. If the sid is ABCDE, then we should send DCBA000E, but I think we send DCBAE000. If you fix that I imagine that the test would pass. Also put a link to that section of the spec into the serialization and deserializarion code

russelltg (changes requested):

fkaa:
Test was failing because `Buf::chunks()` didn't consume anything, adding a `buf.advance(buf.remaining())` fixes it. Also rewrote the parsing to only take into account padding for the last word when parsing, it looks a bit better now as well IMO.

.cargo/config.toml
```diff
@@ -1,2 +1,2 @@
-[target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "linker=clang", "-C", "link-arg=-fuse-ld=lld"]
\ No newline at end of file
+#[target.x86_64-unknown-linux-gnu]
+#rustflags = ["-C", "linker=clang", "-C", "link-arg=-fuse-ld=lld"]
```
russelltg, line 2: Please don't commit this. Feel free to keep this locally (or just install LLD), but it greatly helps incremental test times as it spends most of the time in the linker.

srt-protocol/src/packet/control/srt.rs
```diff
@@ -208,18 +208,28 @@ impl SrtControlPacket {
...
-                    .position(|a| *a != 0)
-                    .unwrap_or(string_bytes.len()); // all null
-
-                string_bytes.resize(string_bytes.len() - shrink_by, 0);
-
-                match String::from_utf8(string_bytes) {
+                // the stream id string is stored as 32-bit little endian words
+                // https://tools.ietf.org/html/draft-sharabayko-mops-srt-01#section-3.2.1.3
+                let mut bytes = Vec::with_capacity(buf.remaining());
+                let mut chunks = buf.chunk().chunks(4).peekable();
```
russelltg, line 214: This is probably fine, but is not technically correct--chunk() doesn't necessarily return all the bytes, as Buf doesn't have to be contiguous (it works because we're constructing it from a contiguous array). Perhaps a cleaner and more correct way to do this would be something along the lines:

```rust
if buf.remaining() % 4 != 0 { return Err(NotEnoughData) }

while buf.remaining > 4 {
     bytes.extend(buf.read_u32_le().to_be_bytes())
}
match buf.read_u32_le().to_be_bytes() {
    // your existing match logic here
}
```

russelltg (changes requested):
It's really close to what I want. I would still love to see a unit test if you have it, but no worries if its' difficult to fetch it. You can see how I have tests from packets from wireshark in raw_handshake_crypto (control.rs).

russelltg:
Also, just a tip, you can just add more commits to your branch, no need to `commit --amend` or whatever you're doing, it makes it so I can see just "new" changes and have a history of what's happened. It'll all get squahsed when I merge. Not a big deal for a small PR like this

.cargo/config.toml
fkaa: I didn't have LLD installed so I temporarily commented it out. Didn't mean to commit it :sweat_smile:

PR #68, opened 2021-01-18
Unbreak rustfmt

lu-zero:


russelltg:
As those test indicate, using those raw strings is not viable (different text, and weird indetation as it's included in the string). I do like the idea of splitting those lines, I'd just create some global consts for the long strings, where there is no indentation. You'd have to update the UI tests for the new error string.

russelltg (changes requested):
Fix error strings

lu-zero:
I started with breaking the lines so `rustfmt` can run. Later I'll check if I can improve the user experience.

Issue #69, opened 2021-01-18
ipv6 support 


lu-zero:
I saw that the current master has at least a TODO about it in `srt-protocol/src/packet/control.rs`:
``` rust
                // TODO: this is probably really wrong, so fix it
                let peer_addr = if ip_buf[4..] == [0; 12][..] {
                    IpAddr::from(Ipv4Addr::new(ip_buf[3], ip_buf[2], ip_buf[1], ip_buf[0]))
                } else {
                    IpAddr::from(ip_buf)
                };
``` 

Is there anything else to be addressed?

russelltg:
Yeah, so I'm pretty sure the "correct" way to do this the packet needs to know if it's being send/received from an ipv6 socket, so it will require a bit of reorganization (right now it just "guesses" by the lower bytes in the ipv6 address)

russelltg:
For anyone who would want to take this issue:

- [ ] Allow packet parsing to know if it's IPV4 or IPV6 (or, alternatively, just store the IP buffer in the packet and rely on the implementation to know), I'm willing to be convinced either way. I'd prefer the first if it's not too messy, to keep the packet structure "pure"
- [ ] Add tests using IPv6

lu-zero:
I can confirm that the it seem so:
``` c++
// in srtcore/core.cpp
    CIPAddress::pton((m_parent->m_SelfAddr), m_piSelfIP, agent.family(), peer);
```

PR #74, opened 2021-01-20
Pass along the socket address kind

lu-zero:
Partially address #69.

### DIFF
--- a/srt-tokio/src/builder.rs
+++ b/srt-tokio/src/builder.rs
@@ -1,4 +1,4 @@
-use std::net::{IpAddr, SocketAddr, ToSocketAddrs};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, ToSocketAddrs};
 use std::{io, time::Duration};
 
 use tokio::net::UdpSocket;
@@ -11,7 +11,7 @@ use crate::{
     crypto::CryptoOptions, multiplex, pending_connection, Packet, PacketCodec, PacketParseError,
     SrtSocket,
 };
-use log::warn;
+use log::{error, warn};
 use srt_protocol::{
     accesscontrol::{AllowAllStreamAcceptor, StreamAcceptor},
     pending_connection::ConnInitSettings,
@@ -58,9 +58,18 @@ use srt_protocol::{
 #[derive(Debug, Clone)]
 #[must_use]
 pub struct SrtSocketBuilder {
-    local_addr: SocketAddr,
+    local_addr: Option<IpAddr>,
+    local_port: u16,
     conn_type: ConnInitMethod,
     init_settings: ConnInitSettings,
+}
+
+fn unspecified(is_ipv4: bool) -> IpAddr {
+    if is_ipv4 {
+        Ipv4Addr::UNSPECIFIED.into()
+    } else {
+        Ipv6Addr::UNSPECIFIED.into()
+    }
 }
 
 /// Describes how this SRT entity will connect to the other.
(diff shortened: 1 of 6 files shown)


srt-tokio/src/builder.rs
```diff
@@ -208,19 +218,28 @@ impl SrtSocketBuilder {
...
                     &mut socket,
                     addr,
-                    self.local_addr.ip(),
+                    local_addr,
                     self.init_settings,
                     sid,
                 )
-                .await?
+                .await;
+                println!("{:?}", r);
```
russelltg, line 231: Remove the println

srt-tokio/src/builder.rs
```diff
@@ -208,19 +218,28 @@ impl SrtSocketBuilder {
                 pending_connection::listen(&mut socket, self.init_settings).await?
             }
             ConnInitMethod::Connect(addr, sid) => {
-                pending_connection::connect(
+                let local_addr = self.local_addr.unwrap_or_else(|| unspecified(addr.ip()));
+                println!("{:?} {:?}", addr, local_addr);
```
russelltg, line 222: Remove this as well. I'd add a `assert!(addr.is_ipv4() == local_addr.is_ipv4())`, to avoid issues with mismatch.

srt-tokio/src/builder.rs
```diff
@@ -81,11 +90,12 @@ pub enum ConnInitMethod {
 }
 
 impl SrtSocketBuilder {
-    /// Defaults to binding to `0.0.0.0:0` (all adaptors, OS assigned port), 50ms latency, and no encryption.
+    /// Defaults to binding to all adaptors, OS assigned port, 50ms latency, and no encryption.
```
russelltg, line 93: Add a note that with listen mode, it defaults to IPV4, and in the docs for the `listen` function

srt-tokio/src/builder.rs
```diff
@@ -238,9 +257,21 @@ impl SrtSocketBuilder {
 
     /// Connects to the remote socket. Resolves when it has been connected successfully.
     pub async fn connect(self) -> Result<SrtSocket, io::Error> {
-        let la = self.local_addr;
+        let addr = match self.conn_type {
+            ConnInitMethod::Connect(addr, _) => addr.ip(),
+            ConnInitMethod::Rendezvous(addr) => addr.ip(),
+            _ => Ipv4Addr::UNSPECIFIED.into(),
```
russelltg, line 263: I think it's a bit cleaner to put the explicit variant, which would also make sure this doesn't compile if more modes are added

srt-tokio/src/builder.rs
```diff
@@ -58,11 +58,20 @@ use srt_protocol::{
...
 #[must_use]
 pub struct SrtSocketBuilder {
-    local_addr: SocketAddr,
+    local_addr: Option<IpAddr>,
+    local_port: u16,
     conn_type: ConnInitMethod,
     init_settings: ConnInitSettings,
 }
 
+fn unspecified(orig: IpAddr) -> IpAddr {
```
russelltg, line 67: Would be cleaner if this just took a `is_ipv4` or `is_ipv6` bool, unless there's a compelling reason that you chose this

russelltg (changes requested):
Looks good! Just a few changes and it'll be good to go.

russelltg:
Also, add a unit test for parsing/serializing a handshake packet with an IPv6 address in it.

srt-tokio/src/builder.rs
lu-zero: probably it should be an explicit error.

lu-zero:
> Also, add a unit test for parsing/serializing a handshake packet with an IPv6 address in it.

Do you have a good way to save packets from the reference implementation?

russelltg:
Yeah, you can grab one from wireshark

srt-tokio/src/builder.rs
russelltg: I think a panic is reasonable, it's a programming error, I think it's fair to just say it's invalid use of the API

PR #82, opened 2021-03-01
Accept Handshake extensions in any order

hexd0t:
The spec doesn't require handshake extensions to be in any specific order.
OBS Studio (using ffmpeg's srt implementation) can be made to send the `streamid` before the `key material`,
e.g. when using the connection string

```
srt://localhost:9710?latency=60000&mss=1464&streamid=#!::u=hexd0t&passphrase=abcdefghijklm&pbkeylen=32&transtype=live
```

Following the robustness principle / Postel's law, allow the extensions to come in any order, and just warn if the flags don't match the actual extensions being sent.

### DIFF
--- a/srt-protocol/src/packet/control.rs
+++ b/srt-protocol/src/packet/control.rs
@@ -498,7 +498,7 @@ impl ControlTypes {
                             0 | 16 | 24 | 32 => crypto_size as u8,
                             c => {
                                 warn!(
-                                    "Unrecognized crypto key length: {}, disabling encryption. Should be 0, 16, 24, or 32 bytes. Disabling crypto.",
+                                    "Unrecognized crypto key length: {}, disabling encryption. Should be 0, 16, 24, or 32 bytes.",
                                     c
                                 );
                                 0
@@ -551,7 +551,8 @@ impl ControlTypes {
                                         if ext_hs != None {
                                             warn!("Handshake contains multiple handshake extensions, only the last will be applied!");
                                         }
-                                        ext_hs = Some(SrtControlPacket::parse(pack_type, &mut buffer)?);
+                                        ext_hs =
+                                            Some(SrtControlPacket::parse(pack_type, &mut buffer)?);
                                     }
                                     3 | 4 => {
                                         if !extensions.contains(ExtFlags::KM) {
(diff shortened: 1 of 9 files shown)


russelltg (changes requested):
Good stuff here. Would love to see a unit test with an example packet (example would be short_ack), as well as the formatting issue that CI is noting.

You can grab a packet from wireshark.

srt-protocol/src/packet/control.rs
```diff
@@ -1526,4 +1522,68 @@ mod test {
             )
         );
     }
+
+    #[test]
+    fn test_unordered_hs_extensions() {
+        //Taken from Wireshark dump of FFMPEG connection handshake
```
russelltg, line 1528: Would be great if you included the full commandline

russelltg (changes requested):

srt-protocol/src/packet/control.rs
hexd0t: This connection was initiated using OBS Studio, which uses ffmpeg via API calls.
The connection string used in OBS was
```
srt://localhost:9710?latency=60000&mss=1464&streamid=#!::u=hex&passphrase=9LS6wkXLpBo&pbkeylen=32&transtype=live
```
(same as in the PR description).
Should I just add that as a comment, or do you need me to recreate this on ffmpeg's cli?

hexd0t: (BTW: No worries about the passphrase, it's just a testing value for local connections)

hexd0t: OBS does start a subprocess for this connection, but using its own ffmpeg cli:
```
obs-ffmpeg-mux.exe" "srt://localhost:9710?latency=60000&mss=1464&streamid=#!::u=hex&passphrase=9LS6wkXLpBo&pbkeylen=32&transtype=live" 1 1 h264 2000 1920 1080 1 1 1 1 30 1 aac "adv_stream_aac" 160 48000 2 "" "" 
```

russelltg: If it's not simple, it's fine how it is. I'll go ahead and merge.

russelltg (approved):

PR #94, opened 2021-04-11
ensure receiver output time is always monotonic

### LINKED ISSUE #93: Decreasing arrival time
My reading of the [`SrtSocket` description](https://docs.rs/srt-tokio/0.2.0/srt_tokio/struct.SrtSocket.html) is that client code can expect that the `Instant` value does not decrease from one yielded value to the next.

I added some logging to my application code

```rust
    let mut last_arrival_time = None;
    while let Some((arrival_time, bytes)) = sock.try_next().await.expect("sock.try_next() failed") {
        if let Some(last_arrival_time) = last_arrival_time {
            if arrival_time < last_arrival_time {
                error!("Out of order packets. last={:?} this={:?}", last_arrival_time, arrival_time);
                continue;
            }
        }
        last_arrival_time = Some(arrival_time);
        ...
    }
```
 and can see that very occasionally the value does decrease, by _10ms_ in this case,

```
[2021-04-07T11:10:03.777845477Z ERROR test::srt] Out of order packets. last=Instant { tv_sec: 4108711, tv_nsec: 501700506 } this=Instant { tv_sec: 4108711, tv_nsec: 501690506 }
```
by _8ms_ in this case,
```
[2021-04-08T06:52:01.946181252Z ERROR test::srt] Out of order packets. last=Instant { tv_sec: 4179629, tv_nsec: 670029779 } this=Instant { tv_sec: 4179629, tv_nsec: 670021779 }
```

Is my reading of the API wrong, or does this represent a problem somewhere?

robertream:
Among other code improvements, this should fix #93.


srt-protocol/src/protocol/mod.rs
```diff
@@ -12,30 +13,42 @@ pub mod stats;
...
-const TIMESTAMP_MASK: u128 = u32::MAX as u128;
-const MAX_DIFF_US: u32 = u32::MAX / 2;
-const WRAP_US: u64 = (u32::MAX as u64) + 1;
-
 #[derive(Copy, Clone, Debug)]
-pub struct TimeBase(Instant);
+pub struct TimeBase {
+    origin_time: Instant,
+    reference_time: Instant,
+    reference_ts: TimeStamp,
```
russelltg, line 27: Would be great if you had some comments here on what the various fields meant

It seems like: origin_time is the old .0
reference_time: is the time of the last wrap? Or something?

srt-protocol/src/protocol/mod.rs
```diff
@@ -152,49 +199,85 @@ impl Sub<TimeSpan> for TimeSpan {
...
+    pub fn instant_from(&self, timestamp: TimeStamp) -> Instant {
+        self.reference_time + (timestamp - self.reference_ts)
     }
 
-    pub fn adjust(&mut self, delta: TimeSpan) {
-        if delta.0 > 0 {
-            self.0 += Duration::from_micros(delta.0 as u64);
-        } else {
-            self.0 -= Duration::from_micros(delta.0.abs() as u64);
+    #[allow(clippy::assign_op_pattern)]
```
russelltg, line 267: This is fine, maybe more correct to just implement the traits.....

srt-protocol/src/protocol/time.rs
```diff
@@ -1,170 +0,0 @@
-use std::cmp::max;
```
russelltg, line 1: Good catch on this file

srt-protocol/src/protocol/receiver/buffer.rs
```diff
@@ -19,7 +19,6 @@ pub struct RecvBuffer {
     head: SeqNumber,
 
     remote_clock: SynchronizedRemoteClock,
-    time_base: TimeBase,
```
russelltg, line 22: Yay! Love to see this not here.

russelltg (changes requested):
Overall, looks great. I'm not sure what a unit test could look like for this.....I guess the best we could do is a synthetic test that makes sure that it rekeys after N time but if it works with the ref impl that's a much better test.

srt-protocol/src/protocol/mod.rs
robertream: Yes, origin_time is .0 or "start_time" in other contexts. I'm open to alternative name suggestions.
The "reference" fields are equivalent time points from their respective time scales that are reference points for mapping between an Instance and a sender TimeStamp.

I'll add this comment to the code.

robertream:
I was going to add a unit test. My thought was to add one on the SynchronizedRemoteClock and/or a parameterized/stochastic test on the RecvBuffer covering general clock stability.

srt-protocol/src/protocol/mod.rs
robertream: Unfortunately I'd have to submit a pull request to the Rust core to do this. The types here are Instant and Duration from std, otherwise I would have certainly implemented the traits as you suggested. :)

robertream:
Ok, I added comments as well as one unit test for monotonicity

russelltg:
Looks great--if you fix the CI failure (looks like some overflow conditions?) i'll merge it

russelltg (changes requested):
Fix CI failure

russelltg:
Great, I was just investigating this and you applied the same fix 👍

robertream:
Ok, why don't you get this past the finish line, I have other matters I need to tend to right now.

PR #102, opened 2021-05-09
major redesign, reprising connection as DuplexConnection

robertream:
@russelltg
It appears that we have started to step on toes, so I'm submitting this mostly finished draft pull request where I have been exploring how to clean up the tokio I/O loop, which lead me down a path of reintroducing a connection struct that this time encapsulates the Sender and Receiver.  I know this is a very large change set, but please consider the benefits from the design direction I am exploring here. I am contradicting my prior advocacy for keeping the sender and receiver algorithms free of communication, concurrent capable. Since we are now just running the whole I/O loop from one thread, concurrency is less important anyhow. I hope you agree that this has significantly simplified the code. I have two different versions of the I/O loop that are alternated somewhat randomly at runtime. One approach uses an Action/Input enum pair and a single function to drive the loop, this is similar to the prior design of the Sender/Receiver enums. The other approach uses a set of command and query methods to drive the loop. I prefer the ergonomics of the the enum approach. It made debug tracing the code trivial too.

One of the larger tests you added recently doesn't compile with all the changes I made, so I commented it out. I don't have the time to go back and fix it, but I can later, I just felt it was important that you take a look at the changes I have made before our paths diverge too far. 

I also fixed several bugs that I ran into while testing. For instance, we weren't respecting SND in the sender algorithm, which broke congestion control. The timestamp_rollover test ran much faster than real time with this bug, but with the fix it no longer does. I have set it to ignored for now. We should replace this long running test with a few more focused unit tests. In fact, it might already be mostly covered by the existing TimeStamp unit tests.

I'd like to finish this up and move onto a first pass at metrics, so quick feedback would be appreciated.


russelltg:
Also nice apparently proptest ICE's the nightly compiler! :partying_face:

robertream:
> I'll take a deeper look at this in a few days. In the meantime--
> 
> All the srt-protocol integration tests are meant to run much faster than realtime, they never check the wall clock they just simulate time passing--so you're saying that now the algorithm takes a full CPU core to run?

The problem is that sending data should be limited by the SND timer, this is enables congestion control to function, but the current implementation can potentially run the sender algorithm every time the next "action" is pulled, regardless of the send timer. Sending should only happen when the clock ticks forward far enough to trigger a SND, but this seems to have added a bit more computational cost when running the faster than real time simulation tests. The algorithm shouldn't take a full CPU to run, the tests just run noticeably slower. This could be due to inefficiencies in the test harnesses/simulation code too. I haven't spent much time optimizing yet.

> Also, I've been using this test to see how well it can deal with higher-bandwidth connections--I should check it in and just set it to be ignored but for now you can grab it with `git restore -s high_bandwidth -- ./srt-tokio/tests/high_bandwidth.rs`. It seems to freeze up now.
>
> I suggest you run it with `cargo test --test=high_bandwidth -- --nocapture`, as it reports statistics. It's not a proper test with pass/fail behavior, it just tries to push a few MB/s through it.

Thanks, that is helpful.

srt-protocol/tests/lossy_deterministic.rs
```diff
@@ -1,200 +1,140 @@
...
-        // (9602806002654919948, 10),
-        (10210281456068034833, 10_000),
+        (3330590297113083014, 1_000),
+        (11174431011217123256, 1_000),
+        (7843866891970470107, 1_000),
+        (940980453060602806, 1_000),
+        (10550053401338237831, 1_000),
+        (9602806002654919948, 1_000),
+        (11134687271549837280, 1_000),
+        (10210281456068034833, 1_000),
```
russelltg, line 31: Are these seeds that failed for you? Or just arbitrary?

srt-tokio/tests/message_splitting.rs
```diff
@@ -12,33 +12,36 @@ const PACKET_SIZE: usize = 1 << 19;
 async fn message_splitting() -> Result<()> {
     let _ = pretty_env_logger::try_init();
 
-    let sender = SrtSocketBuilder::new_connect("127.0.0.1:11124").connect();
+    let sender = SrtSocketBuilder::new_connect("127.0.0.1:11124")
+        .latency(Duration::from_secs(2))
```
russelltg, line 16: Any particular reason for 2s latency here?

srt-protocol/src/connection.rs
```diff
@@ -49,3 +56,378 @@ pub struct ConnectionSettings {
...
+    pub fn ensure_open(&mut self, now: Instant) -> bool {
+        if !self.is_open() {
+            false
+        } else {
+            self.check_timers(now);
+            true
+        }
+    }
+
+    pub fn check_timers(&mut self, now: Instant) -> Instant {
```
russelltg, line 210: We should either have all the check_timers calls be `Instant -> Instant` or `Instant -> ()` instead of some of one some of the other. I think `Instant -> Instant` is the better way to go, as it ensures that you call check_timers and not just next_timer.

srt-protocol/src/connection.rs
```diff
@@ -49,3 +56,378 @@ pub struct ConnectionSettings {
...
+
+    pub fn next_data(&mut self, now: Instant) -> Option<(Instant, Bytes)> {
+        self.receiver.next_data(now)
+    }
+
+    pub fn next_timer(&self, now: Instant) -> Instant {
+        min(self.sender.next_timer(now), self.receiver.next_timer(now))
+    }
+
+    pub fn ensure_open(&mut self, now: Instant) -> bool {
```
russelltg, line 201: The wording "ensure" make me think that this would transition it into an open state, which does not seem like what it does?

srt-protocol/src/connection.rs
```diff
@@ -49,3 +56,378 @@ pub struct ConnectionSettings {
...
+
+        // if status.check_shutdown(now, || sender.is_flushed()) {
+        //     debug!("{:?} sending shutdown", self.settings.local_sockid);
+        //     sender.send_shutdown(now);
+        // }
+
+        self.next_timer(now)
+    }
+
+    pub fn handle_data_input(&mut self, now: Instant, data: Option<(Instant, Bytes)>) {
```
russelltg, line 226: Why is this not just a wrapper over handle_input or vice versa?

srt-protocol/src/connection.rs
```diff
@@ -49,3 +56,378 @@ pub struct ConnectionSettings {
...
+                self.sender.handle_data(item, now);
+            }
+            None => {
+                self.handle_data_stream_close(now);
+            }
+        }
+    }
+
+    pub fn handle_packet_input(&mut self, now: Instant, packet: Option<(Packet, SocketAddr)>) {
+        debug!(
```
russelltg, line 244: Same for this one

srt-protocol/src/protocol/receiver/mod.rs
```diff
@@ -222,22 +175,27 @@ impl Receiver {
...
-            SendControl(packet, self.settings.remote)
-        } else if self.shutdown_flag && self.is_flushed() {
-            Close
-        } else {
-            // 2) Start time bounded UDP receiving. If no packet arrives, go to 1).
-            TimeBoundedReceive(self.next_timer(now))
+        if let Some(timeout) = self.shutdown_flag {
+            if now > timeout {
+                self.lr_ack_acked.1 = self.receive_buffer.next_release()
+            }
```
russelltg, line 181: This is definitely hacky, at least a comment would be nice

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -163,66 +146,74 @@ impl Sender {
...
-            return WaitUntil(self.snd_timer.next_instant());
+        // don't return close until fully flushed
+        let flushed = self.is_flushed();
+        if self.status.check_shutdown(now, || flushed) {
+            debug!("{:?} sending shutdown", self.settings.local_sockid);
+            self.send_control(ControlTypes::Shutdown, now);
         }
+    }
+
+    pub fn next_timer(&self, now: Instant) -> Instant {
```
russelltg, line 190: I think this here is why the sender algorithm now wakes up so much more than it used to: even when there is no data to be sent it requests wakeup to send a packet!

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -233,11 +224,18 @@ impl Sender {
...
+        else if self.transmit_buffer.is_empty() {
             // TODO: the spec for 3) seems to suggest waiting at here for data,
             //       but if execution doesn't jump back to Step1, then many of
             //       the tests don't pass... WAT?
-            return WaitForData;
+
+            if self.status.should_drain() && self.send_buffer.len() == 1 {
+                if let Some(packet) = self.send_buffer.pop() {
+                    self.send_data(packet, now);
+                    self.lr_acked_packet = self.transmit_buffer.next_sequence_number;
```
russelltg, line 235: Again this hack could use at least a comment. Maybe we can think of a better way to handle this?

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -256,13 +254,13 @@ impl Sender {
...
         } else if let Some(p) = self.pop_transmit_buffer() {
             self.send_data(p, now);
-        } else if self.close_requested {
-            // this covers the niche case of dropping the last packet(s)
-            if let Some(dp) = self.send_buffer.front().cloned() {
-                self.send_data(dp, now);
+        } else if self.status.should_drain() && self.send_buffer.len() == 1 {
+            if let Some(packet) = self.send_buffer.pop() {
+                self.send_data(packet, now);
+                self.lr_acked_packet = self.transmit_buffer.next_sequence_number;
```
russelltg, line 263: Again here

srt-protocol/src/protocol/receiver/mod.rs
```diff
@@ -118,7 +108,7 @@ pub struct Receiver {
     receive_buffer: RecvBuffer,
 
     /// Shutdown flag. This is set so when the buffer is flushed, it returns Async::Ready(None)
-    shutdown_flag: bool,
+    shutdown_flag: Option<Instant>,
```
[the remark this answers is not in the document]
robertream: Right, this is a shutdown with a timeout. There were many tests that would fail intermittently/randomly due to injected packet loss or race conditions, this was particularly true of the tokio tests. The connect and rendezvous algorithm tests were failing intermittently for this same reason. We need to implement timeouts for them as well.

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -409,26 +378,15 @@ impl Sender {
         // TODO: reset EXP
     }
 
-    fn handle_handshake_packet(&mut self, handshake: HandshakeControlInfo, now: Instant) {
-        if let Some(control_type) = self.handshake.handle_handshake(&handshake) {
+    pub fn handle_handshake_packet(&mut self, handshake: HandshakeControlInfo, now: Instant) {
+        if let Some(control_type) = self.handshake.handle_handshake(handshake) {
```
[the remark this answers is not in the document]
robertream: Agreed, I was incrementally refactoring, just hadn't gotten to it yet. I also intend to move all the connection Drain/Shutdown/Close logic to DuplexConnection as well, the keepalive timer too, and a few other things.

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -163,66 +146,74 @@ impl Sender {
...
+    }
+
+    pub fn is_open(&self) -> bool {
+        self.status.is_open()
+    }
+
+    fn on_snd(&mut self, now: Instant) {
+        let now_ts = self.transmit_buffer.timestamp_from(now);
+        let window_length = self.metrics.rtt + self.settings.send_tsbpd_latency;
+        self.send_buffer.release_late_packets(now_ts, window_length);
```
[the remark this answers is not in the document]
robertream: Yes, I believe so. I implemented this as a solution to fixing a stall condition while draining on close/shutdown, yet this should be handled by the send_buffer.pop() logic further down in on_snd, so I should either properly implement DropRequest or remove for now.

srt-protocol/src/protocol/sender/mod.rs
robertream: I struggled with this a bit too, the way I see it, there are two options:
1) avoid modifying state directly to arrive at a flush state, but rather have SND responsible for explicitly transitioning to a "flushed" state by updating the status field, so is_flushed would merely check to see if the sender is in a flushed state
2) extract method on this draining flush logic to document the intent of the ugly better

srt-protocol/src/protocol/sender/mod.rs
robertream: I intended to at least remove this redundancy.

srt-protocol/src/connection.rs
robertream: I agree it's awkward. This was aimed towards combining is_open and check_timers. I believe the better design choice is to combine check_timers and next_timer as you pointed out in your previous comment. I was actually going to abandon ensure_open during cleanup. Again, this was an in flight draft.

srt-protocol/src/connection.rs
robertream: I did this in order to differentiate between the two public API designs I was experimenting with. Action/Input vs multiple command and query approach (i.e. is_open, check_timers, handle_packet, handle_data, next_packet, next_data)

srt-protocol/src/connection.rs
robertream: same

srt-protocol/src/protocol/sender/mod.rs
robertream: Good point! :)

robertream: My first pass at fixing this just now reduces the execution time for timestamp_rollover to roughly 7min on my MacBook. I'm still not pleased with this. But it that is promising, so I'll run flamegaph on it and see if there are any low hanging fruit for further optimization.

russelltg: Sounds good! We can probably just run 3 hours of simulation instead of a full 24, as rollover occurs every ~1.2 hours, which should help by almost an order of magnitude.

srt-protocol/tests/helpers/simulation.rs
```diff
@@ -0,0 +1,209 @@
+use bytes::Bytes;
```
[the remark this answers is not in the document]
russelltg: I also like your approach better--I'll add to your PR the removal of mine and porting to yours.

srt-tokio/tests/message_splitting.rs
russelltg: Yeah good question--as messages can potentially be unbounded in size and shutdown should end eventually. I'm not sure that linking it to TSBPD is necessary though--maybe link with rtt? 

Another thing is senders should probably cancel the current message if it receives a shutdown. 

Willing to be convinced here

srt-protocol/src/protocol/sender/mod.rs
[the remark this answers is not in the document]
russelltg: I think we should not have that as part of this PR. Either leave it with a TODO comment and linked issue or remove and leave a TODO with a linked issue.

srt-protocol/tests/helpers/simulation.rs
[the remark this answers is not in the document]
robertream: ok, feel free to also reorganize the module layout if you wish, I don't think we really need both a helper and simulation module

srt-tokio/tests/message_splitting.rs
robertream: I felt that waiting for the TSBD latency duration in order to shutdown allows for the receiver to drain any outstanding packets without losing data or real time playback. I'm fine with canceling current in-flight messages on shutdown, but messages that are waiting for playback should be drained I think. Which requires waiting for roughly TSBD latency

srt-protocol/src/protocol/sender/mod.rs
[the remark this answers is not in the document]
robertream: Ok, now I want to leave this on the Sender for now because that DuplexConnection doesn't have the means for buffering packets to send yet, and maybe it never will. For this PR I'm going to just leave this the way it is. Eventually I will probably explore managing buffers externally using some form of context or dependency injection, perhaps via a "send" Fn, arena allocator, etc.

russelltg:
So things that need to happen before I merge this:

- [ ] Rename and recomment shutdown_flag
- [ ] Comment the missing `DropRequest` implementation
- [ ] Fix `ensure_open` to either be less confusing or remove

As well as any other cleanup you want to do.

Things that don't need to be part of this PR but could be or could be new issues:

- [ ] Research the return on `sender/mod.rs:216`
- [ ] Move handshake returning behavior into `DuplexConnection`
- [ ] `DropRequest`
- [ ] Figure out shutdown + large messages [discussion](https://github.com/russelltg/srt-rs/pull/102#discussion_r629526959) [discussion](https://github.com/russelltg/srt-rs/pull/102#discussion_r629535324)
- [ ] Make `check_timers` calls consistent [discussion](https://github.com/russelltg/srt-rs/pull/102#discussion_r629531399)
- [ ] Figure out public API [discussion](https://github.com/russelltg/srt-rs/pull/102#discussion_r629533363)

russelltg:
I'll fix up the not_enough_latency test

russelltg:
Waiting for CI to run, apparently there's an issue with github actions rn. Will merge if everything looks good

PR #109, opened 2021-05-30
full ack

robertream:
While reading through the spec, researching the implementation of Too Late Drop Packet, I realized that the ACK/ACK2 wasn't using the sequence numbers correctly. They should be using what amounts to a `Wrapping<NonZeroU32>`. I also now understand the difference between a Light, Small, and Full ACK, and we don't implement them correctly. This PR is just a first step in correcting this. I'll add an issue to track this as well. I also want to move all of the primitive types to a common module. I don't care what the name of it is, but there are now several value types used in the protocol, so I called it "values". I'd like to move the other sequence numbers, ID's, and time value types to the same module. They tend to be used in similar places in the code and it could clean up the code nicely. What do you think?

See: https://datatracker.ietf.org/doc/html/draft-sharabayko-mops-srt-00#section-3.2.3

   There are several types of ACK packets:

   *  A Full ACK control packet is sent every 10 ms and has all the
      fields of Figure 10.

   *  A Lite ACK control packet includes only the Last Acknowledged
      Packet Sequence Number field.  The Type-specific Information field
      should be set to 0.

   *  A Small ACK includes the fields up to and including the Available
      Buffer Size field.  The Type-specific Information field should be
      set to 0.

   The sender only acknowledges the receipt of Full ACK packets (see
   ACKACK).

   The Lite ACK and Small ACK packets are used in cases when the
   receiver should acknowledge received data packets more often than
   every 10 ms.  This is usually needed at high data rates.  It is up to
   the receiver to decide the condition and the type of ACK packet to
   send (Lite or Small).  The recommendation is to send a Lite ACK for
   every 64 packets received.


russelltg:
It seems my PR has except yours except yours is a wrapper around u32 and not i32, which is better I think, I'll update my PR to match

With my new AckControlInfo, I think I might be able to get away with making it a NonZeroU32, I'll give it a try. Wish I had seen this earlier today!

russelltg:
Superceded by #111

PR #115, opened 2021-07-05
Fix leaking vectors

### LINKED ISSUE #114: Buffer size is constantly growing while receiving packets (memory leak or... ?).
Hey everyone So i fired the [example](https://github.com/russelltg/srt-rs/blob/master/srt-tokio/examples/receiver.rs) for 1hour 15 minutes. And the problem is the internal vector is constantly growing (with some valgrind examples below) causing more and more memory usage over time.
I used the latest version of the repo - `srt-tokio = {git = "https://github.com/russelltg/srt-rs", branch = "master"}` . The way i received packets: 
1) Run  srt-live-transmit udp://:1234 srt://:4201 -v
2) Run ffmpeg -re -stream_loop -1 -i video.ts -c copy -f mpegts "udp://127.0.0.1:1234?pkt_size=1316"
3) valgrind --tool=massif ./my_proj (copy and paste of the example above) 
Results: 
First project (copy and paste from the [example](https://github.com/russelltg/srt-rs/blob/master/srt-tokio/examples/receiver.rs) ): 

Proj1
Proj1

Second project (there are some ffmpeg allocations involved but they work fine, don't mind them): 

Proj2
Proj2

So there are a couple of questions: 
1) Is such behavior expected ? If so when the `vec` will stop growing ? 
2) Is it useful to use `srt_socket.flush().await.unwrap();` to prevent such behavior ? 
Thanks in advance!

Lighty0410:
Not quite sure if the logic is 100% correct, but at least it works in my project. Faced no leaks with these changes after 1h15m stream so far. 
Probably gonna close #114  
Any suggestions are appreciated.

### DIFF
--- a/srt-protocol/src/connection.rs
+++ b/srt-protocol/src/connection.rs
@@ -261,6 +261,7 @@ impl DuplexConnection {
             self.settings.local_sockid,
             packet
         );
+
         match packet {
             Some(packet) => self.handle_packet(now, packet),
             None => self.handle_socket_close(),
--- a/srt-protocol/src/protocol/receiver/mod.rs
+++ b/srt-protocol/src/protocol/receiver/mod.rs
@@ -37,7 +37,7 @@ struct LossListEntry {
     k: i32,
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 struct AckHistoryEntry {
     /// the highest packet sequence number received that this ACK packet ACKs + 1
     ack_number: SeqNumber,
@@ -48,7 +48,7 @@ struct AckHistoryEntry {
     departure_time: Instant,
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 struct PacketHistoryEntry {
     seqno: SeqNumber,
     time: Instant,
@@ -207,7 +207,7 @@ impl Receiver {
         self.receive_buffer.synchronize_clock(now, ts)
     }
 
-    fn caclculate_receive_rates(&self) -> (u32, u32) {
+    fn caclculate_receive_rates(&mut self) -> (u32, u32) {
         if self.packet_history_window.len() < 16 {
             (0, 0)
         } else {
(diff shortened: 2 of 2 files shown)


russelltg:
The is not the solution I had in mind--but it's definitely a start. If you want to take a stab, the steps would look like:

1. Look at the reference implementation and see how they do it--is it fixed size, configurable, etc. I beleive it is configurable but I'm not totally sure. This would probably need to be exposed in `SrtSocketBuilder` and added to `ConnectionSettings`
2. Change these from `Vec`'s to `Deque`'s--the way you've currently implemented it it's `O(buffer size)` to reallocate, copy, and move into, when there is no real reason for this as Deques can do all these operations in constant time--and every time an element is added to these Deque's we'd need to pop one off if the buffer size max is hit.
3. Make sure there's no spots that are assuming that the full packet history is available, and that we don't need to change any of the logic surrounding these buffers

Lighty0410:
> The is not the solution I had in mind--but it's definitely a start. If you want to take a stab, the steps would look like:
> 
>     1. Look at the reference implementation and see how they do it--is it fixed size, configurable, etc. I beleive it is configurable but I'm not totally sure. This would probably need to be exposed in `SrtSocketBuilder` and added to `ConnectionSettings`
> 
>     2. Change these from `Vec`'s to `Deque`'s--the way you've currently implemented it it's `O(buffer size)` to reallocate and copy, when there is no real reason for this as `Deque`s can do all these operations in constant time--and every time an element is added to these Deque's we'd need to pop one off if the buffer size max is hit.
> 
>     3. Make sure there's no spots that are assuming that the full packet history is available, and that we don't need to change any of the logic surrounding these buffers

Thank you a lot. Gonna explore these cases and push a commit when i'll have time (probably this/next week). Should i close the PR for now though ?

russelltg:
You can just update this PR, no need to close it

russelltg:
These should be all fixed now as of #117. Open an issue if that's not the case

Issue #73, opened 2021-01-18
Implement too-late packet drop


russelltg:
As detailed in 
https://tools.ietf.org/html/draft-sharabayko-mops-srt-00#section-4.6

Previous partial attempt: https://github.com/russelltg/srt-rs/pull/102/commits/c94d2b0a862ad80dbf62f49c02ecbd9584e7bf8f

PR #116, opened 2021-07-17
drop too late packets

robertream:
#73


robertream:
@russelltg I'm not sure why the build is failing on windows and Ubuntu, but I ran the tests locally on MacOS and PopOS (Ubuntu) and the tests are passing

russelltg:
Taking a look at this--might take me a bit--quite a large changeset here.

russelltg:
Sounds good--your contributions have been good so I'm just going to do a brief look and not a long review, as this project is still alpha-stage anyways, things can change and I definitely want these features you are adding and I'm fine with your new structure

srt-protocol/src/protocol/mod.rs
```diff
@@ -164,6 +169,12 @@ impl fmt::Debug for TimeSpan {
     }
 }
 
+impl From<Duration> for TimeSpan {
+    fn from(duration: Duration) -> TimeSpan {
+        TimeSpan::from_micros(duration.as_micros() as i32)
```
russelltg, line 174: Should this panic on overflow? Don't want to create weird situtations

russelltg (changes requested):

russelltg:
Have a few comments from a quick look, but other than that LGTM

russelltg:
These message_splitting errors are concerning, any clue why the're coming up?

srt-tokio/tests/stransmit_interop.rs
```diff
@@ -110,15 +110,15 @@ async fn stransmit_client() -> Result<(), Error> {
 
     // stranmsit process
     let mut stransmit = allow_not_found!(Command::new("srt-live-transmit")
-        .arg("udp://:2002")
-        .arg("srt://127.0.0.1:2003?latency=182")
+        .arg("udp://:2452")
```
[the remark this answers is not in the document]
robertream: When running in parallel there were conflicts with some of the tests, so I just changed them all, to ensure there were no conflicts. Yes, a little OCD, perhaps unnecessary.

srt-protocol/src/seq_number.rs
```diff
@@ -15,3 +15,11 @@ pub fn seq_num_range(begin: SeqNumber, past_end: SeqNumber) -> impl Iterator<Ite
         }
     })
 }
+
+impl SeqNumber {
+    pub fn increment(&mut self) -> Self {
```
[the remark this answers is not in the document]
robertream: Good point

srt-protocol/src/protocol/mod.rs
robertream: I was fine with it panicking, this really can’t ever overflow in practical use.

russelltg: I meant it should panic if there's an overflow, I'm not sure if `as i32` panics on overflow

srt-tokio/tests/stransmit_interop.rs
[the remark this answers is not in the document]
russelltg: no conflicts sounds good to me, didn't realize there were any

srt-protocol/src/protocol/mod.rs
russelltg: Yeah, an explicit panic would be good--want to avoid this situtation:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e81f9381a6549cfbd5a00ff3f4eef5eb

russelltg:
Few observations after looking at some packet dumps:
image

This situation seems non-ideal, we send a keepalive 2ms after sending a packet, isn't it supposed to wait a full second?

robertream:
Agreed on the keep alive scheduling. I’ll add an isolation test for this. It shouldn’t be happening.

I believe the message_splitting errors are an end if stream timing issue, the too late packet drop changes altered the behavior of some of these integration tests. The latest spec for Live SRT connections suggests that messages should be limited to single packet sized messages.

srt-tokio/tests/stransmit_interop.rs
```diff
@@ -356,12 +356,13 @@ async fn test_c_client_interop() -> Result<(), Error> {
 
 #[tokio::test]
 #[cfg(not(target_os = "windows"))]
+#[ignore]
```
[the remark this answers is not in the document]
robertream: I believe this was intermittently failing on me.

robertream:
Ok, I just looked at the keep alive implementation. I am only partially done with redesigning/refactoring this. I introduced an Output  struct, it will eventually be an instance of DuplexConnection, and I intend all output packets to go through this instance. This is where the keep alive timer functionality will live.

There are two potential paths to complete the transition:
1) As I keep pulling parts of the Sender and Receiver apart into smaller testable pieces, they may both disappear entirely and what is left is merged into DuplexConnection, and any code sending a packet will delegate to the Output instance.
2) I will either pass an instance of Output or a packet output closure Fn into any methods on Receiver and Sender that can potentially output packets.

srt-tokio/tests/message_splitting.rs
```diff
@@ -10,6 +10,7 @@ use tokio::time::sleep;
 const PACKET_SIZE: usize = 1 << 19;
 
 #[tokio::test]
+#[ignore]
```
russelltg, line 13: Could we just make the message shorter, like 15 packets worth instead of several hundred?

russelltg:
You're much more in the weeds on this one, but 2. with an output struct seems reasonable

robertream:
2 should have probably been 1, I'd start with 2, even if I might end up at 1

I'll experiment with a smaller message size

The bidirectional interop test is failing on the MacOS build, this is what I'm seeing locally. This is another subtle end of stream timing issue that is weirdly OS specific.

srt-tokio/tests/message_splitting.rs
robertream: I'll experiment with this

robertream:
The test that is now failing wasn't failing before. I think this is just transient.

robertream:
Ok, the smaller message worked. Verify the keep alive packet dump and I think we are good?

russelltg:
> The bidirectional interop test is failing on the MacOS

Which build did it fail on? I didn't see that

russelltg:
I think we have a problem in sender where we don't send an ack2 when we need to? Or maybe sending an ack when you don't needa?

image

russelltg:
We could add a total packet count thing to the `lossy_deterministic` test at some point to catch regressions like this

robertream:
I disabled the bidirectional interop test on for MacOS. I'll revisit this one later. It could be a difference in libsrt versions, perhaps a bug in libsrt and not srt-rs library.

My bet is that it's the Receiver. I did see this before, though I thought I had fixed it. I'll add your suggested assertion to lossy_deterministic.

Thanks for being thorough.

russelltg:
Could be both sides, but no ACK2 is definitely a sender bug

russelltg:
Also, the ACK frequency starts at 100Hz and goes down to 10Hz after a few seconds, if that's helpful

robertream:
You are correct, that's a sender bug. I promptly edited my reply. :)

robertream:
Full ACK period is always 10 ms. The 100Hz frequency is probably light ACK, which is a function of the packet receive rate.

russelltg:
10ms is 100Hz....

robertream:
Ha! Of course it is. I need obviously need some caffeine this morning.

srt-protocol/src/protocol/mod.rs
```diff
@@ -164,6 +169,14 @@ impl fmt::Debug for TimeSpan {
     }
 }
 
+impl From<Duration> for TimeSpan {
+    fn from(duration: Duration) -> TimeSpan {
+        let duration = duration.as_micros();
+        assert!(duration < i32::MAX as u128);
+        TimeSpan::from_micros(duration as i32)
```
russelltg, line 176: Changed my mind, you should use https://doc.rust-lang.org/std/primitive.i32.html#impl-TryFrom%3Ci128%3E

robertream:
The ACK timer is fixed at 10ms but the receiver only replies with a Full ACK when packets have been received. My guess still is that there is a decline in the rate of packets received.


lossy_deterministic is showing ACK2 are being sent, it could still be receiver's fault.
I have other things I need to do today, I'll get this fixed though.

russelltg:
All those ack's are after all data has been sent. You should be able to repro that packet dump, it was just the message_splitting test

robertream:
Ok. As far as I can tell, I fixed all the remaining issues.

Issue #40, opened 2020-06-07
Figure out plan for buffer sizing


russelltg:
Currently, the buffers are fully dynamicly sized and can grow to any arbitrary size. This is a security vulnrability, as we are trusting the peer to not send wild packets. For example. if it sends a packet for a seq number far in the future, it adds loss list entries for _each one_, which generally crashes due to out of memory.

- [ ] packet history window
- [ ] loss list

PR #117, opened 2021-09-08
fix remaining steady memory leaks

robertream:
This should fix the remaining memory leaks from #40 #114

@russelltg  We should create separate issues for any security vulnerabilities identified in the issues. Specifically the one were receiving a packet for a data sequence number far in the future will still cause an out of memory crash. I have a simple idea of how to fix this though, but there are quite a lot of changes here, so it takes a while to get this merged, I'll probably end up fixing it too.


russelltg:
That lossy deterministic failure for 32-bit is interesting: 13858442656353620955. I wonder why there are so many losses...random chance or bad protocol?

russelltg:
There's definitely something wrong in that test, packets that were never dropped by the simulation end up dropped by the receiver.

robertream:
@russelltg The dropped packet messages were because I had changed a log::info to a log::warn. These packets were being dropped before as well, dropping packets were precisely what was being tested. The real failures were just a timing sensitivity issues. Locally the tests were passing. I added a small sleep to a hand full of tests and it fixed this fixed the problem.

russelltg:
No, the test fails with that seed. Add it to the beginning of once failing seeds and you'll see it fail

robertream:
I added it locally and it did not fail, I checked it in and it is not failing in the build either.
Is there any other way I can repro this?

russelltg:
Ah, I see I didn't see your changes--I was testing before them (where it did fail)

I'll give this a review later today

russelltg:
So something that was in the last PR too and was curious on your thoughts--if you enable the `high_bandwidth` test and run it like `cargo test --release --test=high_bandwidth -- --nocapture`, you can see that it gets stuck after a few seconds. According to wireshark, it stops sending data and sends DropRequest packets at ~20kHz, which is pretty intense (I assume that's as fast as it can? Idk)

russelltg:
It also sometimes spams ACK's with microsecond spacing:
image

robertream:
Per the spec, Light ACK gets sent every 64 packets received. Is that what you are seeing?

robertream:
I pushed a fix for high_bandwidth. My guess is that the sender isn't getting destructed until after the receiver, so the sender never tells the receiver to shut down the connection. I added an explicit sender.close().

russelltg:
For a small sample, there was 17540 ack's and 26836 data packets, so it does seem like there are more acks than expected. Also good catch for the high_bandwidth, but I think that line is unreachable so won't matter. But still a good change

robertream:
Ok, I know why the ACK spam is happening. I was only using the last Full ACK SeqNumber to determine when the next Lite ACK should be sent.

robertream:
The problem with high_bandwidth has to do with the congestion control algorithm. If this has only changed recently, it could be due to how RTT is being calculated on the receiver side. I'll look closer at it. This shouldn't' be too hard to fix.

russelltg:
Interesting failure, likely spurious but I'll look at it tomorrow. 

None of this addresses it locking up (FWIW this happened before this PR too--so not really a blocker for this one just something that'd be real nice to get fixed)

robertream:
@russelltg with my last commit I also fixed the Drop Message Request DoS/spam you were seeing

PR #121, opened 2021-09-12
configurable LiveBandwidthMode

robertream:


srt-protocol/src/protocol/sender/congestion_control.rs
```diff
@@ -40,40 +40,92 @@ impl StatsWindow<MessageStats> {
...
+    /// The recommended default value is 1 Gbps. The default value is set only for live streaming.
+    ///
+    /// Note that this static setting is not well-suited to a variable input, like when you change the bitrate on an encoder.
+    /// Each time the input bitrate is configured on the encoder, this value should also be reconfigured.
+    Set(DataRate), // m_llMaxBW != 0
+
+    /// Set the SRT send input rate and overhead.
+    /// In this mode, SRT calculates the maximum bandwidth as follows:
+    ///
+    ///   Maximum Bandwidth = Input Rate * (1 + Overhead / 100)
```
russelltg, line 60: This implies overhead is % (as it is in the reference impl), but below it's marked as a DataRate in bytes per second?

srt-protocol/src/protocol/sender/mod.rs
```diff
@@ -353,7 +353,6 @@ impl Sender {
     }
 
     fn on_snd_event(&mut self, now: Instant) {
-        self.snd_timer.reset(now);
```
russelltg, line 356: Why no more reset?

robertream: SND is a periodic timer. In SRT Live mode the SND period is calculated based on the max bandwidth limit and the average packet size. Resetting the timer here has the effect of lengthening the period which effectively lowers the max bandwidth limit. I was seeing an impact on throughput when I ran the high bandwidth test with extreme bitrates that pushed the limits of the implementation and/or my laptop (e.g. 800Mbit).

russelltg: Cool! Was just curious

PR #127, opened 2021-10-18
implement buffer receive sizing

### LINKED ISSUE #126: The receiver stops receiving data unexpectedly.
Hey everyone! I have a remote srt-live-transmit service i try to get data from to my local machine. However, i cannot. 
For some weird reasons the receiver randomly stops working. And the problem is i can't make head nor tail of the issue. 
There's around ~50ms ping on average between me and the remote service.  I tried different methods to eliminate the issue:
1) Different  srt-live-transmit settings. 
2) Different SrtSocketBuilder settings. 
3) Different server settings/location/hardware (within all the same ~50ms). 
What's happening: 
1) I start the srt-live-transmit service on the remote machine - `srt-live-transmit -loglevel:debug udp://:1234 srt://:12345latency=1000&oheadbw=100`
2) Use ffmpeg to push data to it - `ffmpeg -re -stream_loop -1 -i stream.ts -c copy -map 0 -f mpegts "udp://127.0.0.1:1234?pkt_size=1316"`
3) Run example from here - https://github.com/russelltg/srt-rs/blob/master/srt-tokio/examples/receiver.rs with the appropriate addr/port. 
4) The receiver just randomly stops getting packets. However, the connection remains active for quite a while (maybe a minute or two). Furthermore the connection also remains active on the server side! And after a minute or two it drops on both sides. 

And there a couple of weird stuff is going on: 
1) The old version available on crates.io seems to work fine. There's a memory leak, but it works. 
2) Original implementation/ffplay srt also works fine. 
3) There are no issues if i run the srt-live-transmit  locally and connect to it locally. 

There's a trace log (i can replace it if needed with the full version, but it's pretty huge) - https://gist.github.com/Lighty0410/a4cdacef1b37599fee85c5888c0dab52
Thanks in advance!

russelltg:
closes  #126 (@Lighty0410 could you confirm that this fixes the issue?)

@robertream It would be fantastic if you had a minute to review this. I've invited you to be a collaborator on the repo so you can be assigned as a reviewer. I would appreciate it if you still opened PRs for changes that are non-trivial (which I also do).

It also cleans up some sequence number range stuff, making them all exclusive ranges (except the ones in DropRequest, which are defined by the std to be inclusive)

### DIFF
--- a/srt-protocol/src/protocol/receiver/buffer.rs
+++ b/srt-protocol/src/protocol/receiver/buffer.rs
@@ -947,4 +947,37 @@ mod receive_buffer {
 
         assert_eq!(buf.buffer_available(), 0);
     }
+
+    #[test]
+    fn wrong_lrsn_after_drop_all() {
+        let tsbpd = Duration::from_secs(2);
+        let start = Instant::now();
+        let init_seq_num = SeqNumber(5);
+
+        let mut buf = ReceiveBuffer::new(start, tsbpd, init_seq_num, 8192);
+
+        let now = start;
+        let _ = buf.push_packet(
+            now,
+            DataPacket {
+                seq_number: init_seq_num + 3,
+                payload: b"yas"[..].into(),
+                ..basic_pack()
+            },
+        );
+
+        assert_eq!(buf.next_ack_dsn(), init_seq_num);
+
+        // pop_next_message is strange, may want some cleanup
+        assert_eq!(
+            buf.pop_next_message(now + tsbpd + Duration::from_millis(10)),
+            None
+        );
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


robertream:
> @Lighty0410 more testing is always appreciated, but I'm not that worried
> 
> @robertream
> 
> > Don’t use both Range and RangeInclusive, everywhere in the code that is related to DropRequest, including the new_drop_request function, just use RangeInclusive.
> 
> I kinda think we should stick with Range where we can--rust uses half-open ranges almost everywhere, and they clean up the code quite a bit, and is assumped most places in SRT spec (except for here, I guess). I feel like it's the more consistent thing to do

I'm fine with that. This isn't a hill I'd die on.

Issue #104, opened 2021-05-16
Make DuplexConnection responsible for returning handshakes


russelltg:
Should be easy, will require thinking about how DuplexConnection sends packets.

robertream:
I might agree, but why do you feel this is important? Is it a matter of design esthetics?

Regarding how packets are sent, I was expecting to eventually revisit this, but I want to avoid reintroducing async/await into the protocol code, so absent  something like generalized co-routines or effect types becoming available in the language. I suspect that the only way to do this then, is to invert control over the send buffers. This would also have the beneficial effect of getting us one step closer to a potential zero/low allocation implementation suitable for no_std scenarios.

Issue #70, opened 2021-01-18
statistic reporting


lu-zero:
If https://github.com/russelltg/srt-rs/issues/53 wants to be more or less compatible with the reference implementation API, it might be good to support extracting the same statistics.

``` c
// Performance monitor with Byte counters for better bitrate estimation.
int srt_bstats(SRTSOCKET u, SRT_TRACEBSTATS * perf, int clear);

// Performance monitor with Byte counters and instantaneous stats instead of moving averages for Snd/Rcvbuffer sizes.
int srt_bistats(SRTSOCKET u, SRT_TRACEBSTATS * perf, int clear, int instantaneous);
```

With `SRT_TRACEBSTATS` containing the following:

| Statistic                                           | Type of Statistic | Unit of Measurement | Available for Sender | Available for Receiver | Data Type | Completed                |
|-----------------------------------------------------|-------------------|---------------------|----------------------|------------------------|-----------|--------------------------|
| [msTimeStamp](#msTimeStamp)                         | accumulated       | ms (milliseconds)   | ✓                    | ✓                      | int64_t   | <ul><li>- [ ] </li></ul> |
| [pktSentTotal](#pktSentTotal)                       | accumulated       | packets             | ✓                    | -                      | int64_t   | <ul><li>- [ ] </li></ul> |
| [pktRecvTotal](#pktRecvTotal)                       | accumulated       | packets             | -                    | ✓                      | int64_t   | <ul><li>- [ ] </li></ul> |
| [pktSentUniqueTotal](#pktSentUniqueTotal)           | accumulated       | packets             | ✓                    | -                      | int64_t   | <ul><li>- [ ] </li></ul> |
| [pktRecvUniqueTotal](#pktRecvUniqueTotal)           | accumulated       | packets             | -                    | ✓                      | int64_t   | <ul><li>- [ ] </li></ul> |
| [pktSndLossTotal](#pktSndLossTotal)                 | accumulated       | packets             | ✓                    | -                      | int32_t   | <ul><li>- [ ] </li></ul> |
| [pktRcvLossTotal](#pktRcvLossTotal)                 | accumulated       | packets             | -                    | ✓                      | int32_t   | <ul><li>- [ ] </li></ul> |
| [pktRetransTotal](#pktRetransTotal)                 | accumulated       | packets             | ✓                    | -                      | int32_t   | <ul><li>- [ ] </li></ul> |
[...]

PR #130, opened 2021-11-18
statistics

### LINKED ISSUE #106: Clean up check_timers functions to have consistent signatures
I think they should have `Instant -> Instant` signatures, returning the next wakeup time.

robertream:
@russelltg this is mostly to address #70, but it also includes yet another major redesign of the srt_protocol module. The redesign made it easier for me to both test and implement statistics, but I also fixed a few bugs on the way as I added more to the unit test coverage. I also fixed #106 and #104. Sorry, I know it might be an overwhelming PR to review, but it would be great to get it merged so we can iteratively fill out all the desired statistics with much smaller commits. :)


srt-tokio/src/socket.rs
```diff
@@ -201,6 +222,14 @@ impl SrtSocket {
     pub fn settings(&self) -> &ConnectionSettings {
         &self.settings
     }
+
+    pub fn stats(&mut self) -> &mut impl Stream<Item = SocketStatistics> {
```
russelltg, line 226: The reference implementation also has a way to clear statistics--which I believe is the default when you get the statistics. We should have some sort of mechanism for that

srt-protocol/src/connection/mod.rs
```diff
@@ -188,55 +141,69 @@ impl DuplexConnection {
...
-        packet
+        self.output.pop_packet().map(|p| {
+            self.timers.reset_keepalive(now);
+            self.statistics.tx_all_packets += 1;
+            // payload length + (20 bytes IPv4 + 8 bytes UDP + 16 bytes SRT)
+            match &p {
+                Packet::Data(d) => {
+                    self.statistics.tx_data += 1;
+                    // payload length + (20 bytes IPv4 + 8 bytes UDP + 16 bytes SRT)
+                    self.statistics.tx_bytes += d.payload.len() as u64 + 44;
```
russelltg, line 165: Would prefer all these to be const globals

srt-protocol/src/connection/mod.rs
```diff
@@ -355,27 +338,74 @@ impl DuplexConnection {
...
+        ReceiverContext::new(
+            &mut self.timers,
+            &mut self.output,
+            &mut self.statistics,
+            &mut self.receiver,
+        )
+    }
+
+    fn debug(&self, tag: &str, now: Instant, debug: &impl Debug) {
+        self.log(log::Level::Debug, tag, now, debug);
```
russelltg, line 379: Just a nitpick, but this should probably use the macro directly so you can compile various log levels out of the program

srt-protocol/src/protocol/time/base.rs
```diff
@@ -0,0 +1,135 @@
...
+impl TimeBase {
+    pub fn new(start_time: Instant) -> Self {
+        Self {
+            origin_time: start_time,
+            reference_time: start_time,
+            reference_ts: TimeStamp::MIN,
+        }
+    }
+
+    #[allow(clippy::trivially_copy_pass_by_ref)]
```
russelltg, line 27: Why suppress this? Can't we just pass by `self`?

srt-protocol/src/protocol/receiver/buffer.rs
```diff
@@ -158,6 +164,19 @@ impl MessagePacketCount {
     }
 }
 
+#[derive(Debug, Default)]
+pub struct ReceiveBufferStatistics {
```
russelltg, line 168: I don't see this used anywhere??

srt-protocol/src/protocol/receiver/buffer.rs
```diff
@@ -218,92 +241,89 @@ impl ReceiveBuffer {
...
-    fn index_for_seqno(&self, s: SeqNumber) -> Option<usize> {
-        if s < self.seqno0 {
+    fn index_for_seqno(&self, seq_number: SeqNumber) -> Option<usize> {
+        if seq_number < self.seqno0 {
             None
         } else {
-            Some(usize::try_from(s - self.seqno0).unwrap())
+            let index = usize::try_from(seq_number - self.seqno0).unwrap();
+            //            assert!(index < self.buffer.len(), "{:?} !< {:?}", index, self.buffer.len());
+            Some(index)
```
russelltg, line 250: This can probably be rolled back--I assume you found that that assert was not true?

russelltg (commented):
It would be great to see statistics added to `srt-transmit`. For now it would be fine to get print the `Debug` implemenation

Also--remind me why we have both `run_handler_loop` and `run_input_loop`?

I'm a bit unsure about the new package structure of "export everything"--i'm much more generally in favor of not exporting private types (I noticed this with `ReceiverContext`, which seems clearly private, but it is exported from the crate...). I'm okay with it for now, at some point I'll make sure our exports are minimal.

Overall looks okay though, some of it I didn't fully look over because it's impossible to see what actually changed in these files...but we have a great testsuite which makes me much more confident in general.

russelltg:
Also, I'm sure you saw but 

```rust
warning: field is never read: `adjustments`
  --> srt-protocol/src/protocol/receiver/time.rs:16:5
   |
16 |     adjustments: u64,
   |     ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `drift`
  --> srt-protocol/src/protocol/receiver/time.rs:17:5
   |
17 |     drift: (TimeSpan, TimeSpan),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `srt-protocol` (lib) generated 2 warnings
```

srt-protocol/src/statistics.rs
```diff
@@ -0,0 +1,494 @@
+use std::time::Duration;
+
+/// SRT provides a powerful set of statistical data on a socket. This data can be used to keep an eye
+/// on a socket's health and track faulty behavior.
+///
+/// Statistics are calculated independently on each side (receiver and sender) and are not exchanged
+/// between peers unless explicitly stated.
+#[derive(Debug, Eq, PartialEq, Default, Clone)]
+pub struct SocketStatistics {
```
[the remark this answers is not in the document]
robertream: good point

srt-tokio/src/socket.rs
```diff
@@ -201,6 +222,14 @@ impl SrtSocket {
...
+        &mut self.statistics
+    }
+
+    pub fn statistics(&mut self) -> &mut impl Stream<Item = SocketStatistics> {
+        &mut self.statistics
+    }
```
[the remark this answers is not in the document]
robertream: since the polling of statistics can interfere with the packet send rate, I decided to make the 

> Don't think we need two of these....
> 
> Anyways, I think this means that you can't get statistics when you use send_all, which is the most canonical way to use `SrtSocket`... (can't make to `&mut self` references)
> 
> I would probably prefer if there was a `pub fn statistics(&self) -> SrtSocketStatistics`, and that handle had a `read_and_clear(&self)`, `read(&self)` as well as a `clear_statistics(&self)` function, which allows you to poll statistics at any frequency, clear whenever you want, as well as atomically do a read+clear which is probably the most typical usecase.

I was imagining providing a function that splits the streams (input, output, statistics, and possibly even errors/events?), so that they can be read from separately, potentially even from separate tasks. This can address the `send_all` scenario you mention, but otherwise also lets you use `select!` and just call specific read methods on the socket directly.

I am not married to the current design. The main motivation I had was to avoid the possibility of polling at too high a frequency, which could impact throughput in high bandwidth scenarios.

srt-protocol/src/protocol/time/base.rs
robertream: I think this was copied from elsewhere, I am unsure why this was here, I imagine just to appease clippy at some point.

srt-protocol/src/protocol/receiver/buffer.rs
robertream: This is dead code, left over from an early iteration on the design, I'll delete.

srt-protocol/src/connection/mod.rs
robertream: Fair, good point. One goal to normalize the log formatting approach using structured logging of some kind.

srt-tokio/src/socket.rs
robertream: This can be handled as part of a separate issue, specifically one dealing with a C API. My main goal was to just get something working so we could iterate on the quality of the metrics while we figure out the ergonomics of the API.

srt-protocol/src/protocol/receiver/buffer.rs
robertream: Yes, and I revisited the code, in a type system as powerful as Rust's is, I find defensive runtime assertions in production code to be a code smell, often times a result of Primitive Obsession.

srt-protocol/src/connection/mod.rs
robertream: I agree, yet I think we can do this later, when we better understand where it should go. I don't think we're done with this part of the code.

srt-protocol/src/connection/mod.rs
russelltg: Fair enough--it's fine for now

robertream:
> It would be great to see statistics added to `srt-transmit`. For now it would be fine to get print the `Debug` implemenation

Yes, I was going to do this next. 
 
> Also--remind me why we have both `run_handler_loop` and `run_input_loop`?

I wasn't sure which API I liked best yet. We could also keep both, but extract the run_input_loop into a separate struct that wraps the DuplexConnection. This would reduce the confusion.

> I'm a bit unsure about the new package structure of "export everything"--i'm much more generally in favor of not exporting private types (I noticed this with `ReceiverContext`, which seems clearly private, but it is exported from the crate...). I'm okay with it for now, at some point I'll make sure our exports are minimal.

I agree, and all of the `protocol` module is really just private. In this process of redesign/refactoring it was helpful to keep most everything as public. If we organize the modules so the public surface area and the private surface area are divided close to the root of the library, the visibility of the "internal" private modules is not a big concern. I think `packet`, `connection`, and `settings` will likely be the public surface area, the rest will be fully private, including `sender` and `receiver`.

ReceiverContext and SenderContext are used to invoke/implement behavior (i.e. mutations) for the Receiver and Sender "objects", yet implementing these behaviors in a way that is intimately familiar and responsible with the data of the Receiver and Sender. The design of ReceiverContext and SenderContext is influenced by (or at least my best attempt at) the [DCI](https://en.wikipedia.org/wiki/Data,_context_and_interaction) design pattern.

I chose this approach because dependency injection is problematic in general in most languages, but especially in Rust. Having used Go quite a bit at work the last few years, I also dislike passing dependencies into functions everywhere, it makes the call site messy. We already do this with "now", but ultimately I was seeing the need for multiple objects that would be participating in the interaction that implements these behaviors (packet output, metrics, timers, errors/events, logging, etc.). So it started to look like DCI could help.

> Overall looks okay though, some of it I didn't fully look over because it's impossible to see what actually changed in these files...but we have a great testsuite which makes me much more confident in general.

I *LOVE* this test suite!

srt-tokio/src/socket.rs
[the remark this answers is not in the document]
robertream: > I would probably prefer if there was a `pub fn statistics(&self) -> SrtSocketStatistics`, and that handle had a `read_and_clear(&self)`, `read(&self)` as well as a `clear_statistics(&self)` function, which allows you to poll statistics at any frequency, clear whenever you want, as well as atomically do a read+clear which is probably the most typical usecase.

Ok @russelltg with the last commit `statistics()` now returns a `&mut impl Stream<Item=SocketStatistics> + Clone`. If you want to access this stream from a separate task you can now `clone()` the statistics stream.

robertream:
> It would be great to see statistics added to `srt-transmit`. For now it would be fine to get print the `Debug` implementation.

@russelltg I added statistics to the `high_bandwidth` test. I made an attempt at `srt-transmit` but it really is quite a rat's nest of code. I don't have the energy to deal with something like that right now, because it would be too tempting for me to dramatically clean up the code which would end up taking much longer than needed to just get it working. I'll defer this exercise to someone else for now, but I imagine I'll eventually circle back and clean up `srt-transmit` too.

srt-tokio/src/socket.rs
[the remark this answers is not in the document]
russelltg: That seems like a reasonable starting point--it's simple and operable. Further improvements to the API can come, maybe with some research on how other libraries implement it.

russelltg:
Yeah...srt-transmit does a bunch of stuff and it's pretty complex. Let's add it when we feel more confident about the interface.

russelltg:
Oops

PR #132, opened 2021-11-20
make connection !Clone

russelltg:


russelltg:
@robertream Do you think this is reasonable? My thought process was that Connections are logically only consumable once, so we should disallow cloning it

robertream:
> @robertream Do you think this is reasonable? My thought process was that Connections are logically only consumable once, so we should disallow cloning it

The type nerd and domain modeling nerd in me enjoys leveraging Rust's type system, including uniqueness/affine type functionality, for much good. I think this is a reasonable justification for disallowing cloning, though I'd accommodating cloning and copying on the constituent components of Connection. I also think there is much need to clean up the whole connection namespace. There's maybe 4-5 structs that carry connection settings around. I'm wanting to build a differentiation in the model/language in the code to where there is a well defined "public" configuration model/API and then there is a defined though somewhat intersecting internal "initialization"/"settings" model for transitioning between the connection negotiation and connection processing phases. I'll demonstrate some of this in my next PR.

robertream:
> > @robertream Do you think this is reasonable? My thought process was that Connections are logically only consumable once, so we should disallow cloning it
> 
> The type nerd and domain modeling nerd in me enjoys leveraging Rust's type system, including uniqueness/affine type functionality, for much good. I think this is a reasonable justification for disallowing cloning, though I'd still what to accommodate cloning and copying on the constituent components of Connection. I also think there is much need to clean up the whole connection namespace. There's maybe 4-5 structs that carry connection settings around. I'm wanting to build a differentiation in the model/language in the code to where there is a well defined "public" configuration model/API and then there is a defined though somewhat intersecting internal "initialization"/"settings" model for transitioning between the connection negotiation and connection processing phases. I'll demonstrate some of this in my next PR.

PR #133, opened 2021-11-21
significant encryption redesign

robertream:
@russelltg I ran out of time with this today, but all tests are passing and it's in a good enough spot to check in. I intended to finish up the Initiator implementation of the Key Material Refresh algorithm, but I still haven't yet figured out how this is supposed to work with duplex/bi-directional connections. Do both sides initiate refresh? If only one side, how do they coordinate which is the active key?


russelltg:
I'm sure I'm missing something, but what's the point of the "upgrading dependencies" change? The way I had it it just selects the newset version that's semver compatiable, so how does this change dependency resolution?

russelltg:
Also, to answer your question it seems like each direction has it's own crypto context entirely:
https://github.com/Haivision/srt/blob/5f3cd06c49a99b74fc7877f9b8c3c3ff29f7507a/srtcore/crypto.h#L82

(I'm sure there's a better reference than that)

russelltg:
Also, I added a simple bidirectional test that you can play around with. It currently passes because there is no way to configure key refresh period, but that should be helpful. I put a TODO in there on where it would need to be configured for the test to be proper

russelltg:
In wireshark you can see if it's properly using the new key: right now packets sent from the rust endpoint are all encrypted with the even key, but it should alternate every 128 packets:

image

robertream:
> I'm sure I'm missing something, but what's the point of the "upgrading dependencies" change? The way I had it it just selects the newset version that's semver compatiable, so how does this change dependency resolution?

Yeah, ok, you can roll it back if you feel like it. I didn't realize how it was configured, feel free to roll that back

robertream:
> Also, to answer your question it seems like each direction has it's own crypto context entirely: https://github.com/Haivision/srt/blob/5f3cd06c49a99b74fc7877f9b8c3c3ff29f7507a/srtcore/crypto.h#L82
> 
> (I'm sure there's a better reference than that)

Ok, I'll have to move the cipher back to Sender and Receiver. I'll also probably split Cipher into Encryption and Decryption respectively.

russelltg:
(Found a better source for it https://github.com/Haivision/srt/blob/master/docs/API/API-socket-options.md#srto_passphrase)

russelltg:
Hmm, there's definitely some problem with that test--it crashes (segfault) sometimes....lovely

russelltg:
Did you mean to delete the test I added?

robertream:
No, I didn’t

russelltg:
That test now hangs after sending 28 packets

robertream:
Hmmm, I wonder if this is due to the rotation code. I'm developing on a mac
and the test is disabled for macos, so I haven't been running it.

russelltg:
You should be able to enable it so long you have libsrt.dylib installed
somewhere it can find it, not that I've tested it at all

russelltg:
@robertream I think I fixed the preannounce, and the test now passes. Sorry if i was stepping on your toes, but I figured since I had a good repro I would go ahead and fix it

robertream:
> @robertream I think I fixed the preannounce, and the test now passes. Sorry if i was stepping on your toes, but I figured since I had a good repro I would go ahead and fix it

I much appreciate it, I had to put this down and am focused on other work currently, no toes were stepped on.

srt-protocol/src/connection/mod.rs
```diff
@@ -371,35 +370,35 @@ impl DuplexConnection {
...
             debug
         );
     }
 
-    fn info(&self, tag: &str, now: Instant, debug: &impl Debug) {
+    fn info(&self, tag: &str, debug: &impl Debug) {
         log::info!(
             "{:?}|{:?}|{} - {:?}",
-            TimeSpan::from_interval(self.settings.socket_start_time, now),
+            TimeSpan::from_interval(self.settings.socket_start_time, Instant::now()),
```
russelltg, line 391: I don't have a huge stake in this, but I would probably prefer this keeps with passing `now` for when we run the connection at non-realtime

srt-tokio/tests/stransmit_interop.rs
```diff
@@ -154,6 +154,19 @@ async fn stransmit_client() -> Result<(), Error> {
     listener.await;
     stransmit.wait()?;
 
+    if time::Instant::now().elapsed() > Duration::MAX {
```
russelltg, line 157: I still don't quite understand this. How is this ever true? This seems like just a roundabout way to disable some code?

russelltg (commented):
I see you added an empty `options/mod.rs` file, seems like this can be removed?

srt-tokio/tests/stransmit_interop.rs
robertream: Yeah, this can be removed. These tests weren't running correctly on the mac, but the compiler was complaining about dead code. You'll probably also have to disable the test for mac os though. I thought I added a comment to the code for this, but obviously didn't...

srt-protocol/src/connection/mod.rs
robertream: ah, good point

srt-tokio/tests/stransmit_interop.rs
russelltg: We can probably add a `#[allow(dead_code)]`, unless it is many many things that would need marking

robertream:
> I see you added an empty `options/mod.rs` file, seems like this can be removed?

Yeah, this wasn't supposed to get in there, I was working on a spike for flushing out a new model for configuration that maps directly to the SRT options, with references and links to the the original and documentation comments mostly copied directly over so it's easier to keep in sink. Basically "options" will be for connection negotiation configuration, including full configuration validation, and "settings" will be the result of the negotiation process, i.e. a fully established connection, basically ConnectionSettings, which will likely include data structures directly copied from the "options" but combined with important settings derived from the negotiation process, e.g. negotiated latency, MMS, and canned handshake response. I can send a draft PR today for you to look at.

srt-tokio/tests/stransmit_interop.rs
robertream: Well, yes, I started there, but the complication is that there are multiple things that would need to be annotated, including static const, struct, and fn. I probably could have figured it all out, I was being expedient. Perhaps all the linux specific non-test code should be in a submodule, which can be annotated accordingly too?

russelltg: Well the *ideal* solution would be to get that test to work on macOS. We do install srt on mac via brew, so there's nothing technically in the way--since you have a mac it would be great if you could investigate it? (Would be another PR, and we could leave this as-is for now)

robertream:
Ok

srt-tokio/tests/stransmit_interop.rs
robertream: Ok, I can do that.

PR #135, opened 2021-11-24
options

robertream:


robertream:
@russelltg The documentation comments need cleanup, clippy complains, but this is basically the idea. Next I was going to add a Validation trait that each options struct, which will implement validation on each individual part of the configuration. In case there are validations that require more context it will be added to the struct's validation at the appropriate level of the hierarchy. There will be a Validated<T> struct that will encode each valid instance, with the result being a Validated<SocketOptions>. Any public connection negotiation functions will only accept Validated<SocketOptions>. This will leverage the compiler to enforce configuration constraints.

robertream:
> @russelltg The documentation comments need cleanup, clippy complains, but this is basically the idea. Next I was going to add a Validation trait for each options struct, which will implement validation on each individual part of the configuration. In case there are validations that require more context it will be added to the struct's validation at the appropriate level of the hierarchy. There will be a Validated struct that will encode each valid instance, with the result being a Validated. Any public connection negotiation functions will only accept Validated. This will leverage the compiler to enforce configuration constraints.

robertream:
@russelltg This is ready for serious review. I'm comfortable with merging this. I'm not done yet, but this is a solid first step in implementing a new configuration API. I imitate the language of the reference library, so configuration is done through options. Everything under the options module are the types used for specifying and validating configurations. You can chose to use a builder to configure a socket or you can use the options data structures. The .set and .with .with2 methods on the builder and options structs are exploratory. I'm not sure so far what is going to stick around, but I like that these builder methods accommodate accessing all the configuration options without having to break out of the context of the builder. With this, we don't need to add every single configuration option to the builder, the more obscure, less used ones can be accessed via .set or .with 

I anticipate that by having all the configuration options exposed as a coherent simple data model, we can also better interface with programs that wish to accommodate arbitrary complex configurations (e.g. cli tools, GStreamer, FFMpeg). Perhaps support serde serialization might even be desirable?

I'd like to get this merged before I finish up SrtListener, because I intend to take use this same approach with the configuration, and so it will depend on this same data model.

russelltg:
Sounds great--I'll give it a review later today. May take me a few days, this is a pretty big changeset!

robertream:
I merged in the latest listener code too. I can just close the other PR. I think I have both in a good state now.

russelltg:
Sounds good. I've been super busy the past few days, hopefully will get around to it soon.

aegroto:
Looking forward for this PR to merged, this crate is blessful for Rust application communicating over SRT but some configuration capabilities may be useful.

srt-protocol/src/listener/session.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug)]
+pub enum SessionState {
+    Pending(Listen),
+    Rejecting(Option<(Packet, SocketAddr)>),
+    Dropping,
+    Open,
+}
+
+impl SessionState {
+    pub fn new_oending(settings: ConnInitSettings) -> Self {
```
russelltg, line 18: Is this supposed to be `new_pending`?

srt-protocol/src/listener/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    fn handle_packet<'s, 'a>(&'s mut self, now: Instant, packet: (Packet, SocketAddr)) -> Action<'a>
+    where
+        'a: 's,
+    {
+        self.stats.rx_packets += 1;
+        //self.stats.rx_bytes += packet
+        let session_id = SessionId(packet.1, packet.0.dest_sockid());
+        let session = match self.sessions.get_mut(&session_id) {
```
russelltg, line 96: This should use the Entry api, will be cleaner and more performant

srt-protocol/src/listener/statistics.rs
```diff
@@ -0,0 +1,19 @@
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
```
russelltg, line 1: Would be good to have these commented at some point, but not a huge deal for this PR

srt-protocol/src/listener/statistics.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub tx_bytes: u64,
+
+    pub delegated_packets: u64,
+    pub delegated_bytes: u64,
+
+    pub cx_inbound: u64,
+    pub cx_opened: u64,
+    pub cx_dropped: u64,
+    pub cx_rejected: u64,
+    pub cx_accepted: u64,
```
russelltg, line 18: cx_?

srt-protocol/src/options/validation.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    fn is_valid(&self) -> Result<(), Self::Error>;
+
+    fn try_validate(self) -> Result<Valid<Self>, Self::Error> {
+        self.is_valid()?;
+        Ok(Valid(self))
+    }
+}
+
+pub trait CompositeValidation: Validation {
```
russelltg, line 14: Would be great to have some comments on what these traits semantically mean. Composite of what, exactly?

srt-protocol/src/options/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+// SRTO_REUSEADDR - socket specific
+// SRTO_STATE - socket specific
+// SRTO_RCVSYN - not even relevant for tokio
+// SRTO_RCVTIMEO - not even relevant for tokio
+// SRTO_SNDSYN - not even relevant for tokio
+// SRTO_SNDTIMEO - not even relevant for tokio
+// SRTO_KMSTATE - legacy, not supporting
+// SRTO_ISN - only good for development scenarios
+// SRTO_UDP_SNDBUF - not really relevant for tokio is it?
+// SRTO_UDP_RCVBUF - not really relevant for tokio, is it?
```
russelltg, line 76: SNDBUF + RECVBUF are really important options for video streaming, and we should definitely expose them eventually

srt-protocol/src/packet/mod.rs
```diff
@@ -138,4 +135,53 @@ impl Debug for Packet {
...
+
+impl fmt::Display for ReceivePacketError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        <Self as fmt::Debug>::fmt(self, f)
+    }
+}
+
+impl std::error::Error for ReceivePacketError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        if let ReceivePacketError::Io(e) = self {
```
russelltg, line 164: I beleive we can always propagate, even for parse errors?

srt-protocol/src/packet/mod.rs
```diff
@@ -138,4 +135,53 @@ impl Debug for Packet {
...
+
+impl From<PacketParseError> for ReceivePacketError {
+    fn from(error: PacketParseError) -> Self {
+        ReceivePacketError::Parse(error)
+    }
+}
+
+impl fmt::Display for ReceivePacketError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        <Self as fmt::Debug>::fmt(self, f)
```
russelltg, line 158: Not crazy about this--this should probably just call display on it's child?

srt-protocol/src/options/encryption.rs
```diff
@@ -0,0 +1,279 @@
...
+impl Debug for Passphrase {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Passphrase").finish()
+    }
+}
+
+// https://github.com/Haivision/srt/blob/master/docs/API/API-socket-options.md#srto_pbkeylen
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum KeySize {
+    Unspecified,
```
russelltg, line 206: Why do we need Unspecified?

srt-protocol/src/protocol/pending_connection/hsv5.rs
```diff
@@ -33,13 +32,36 @@ pub fn gen_hsv5_response(
         }
     };
 
+    let mut acceptor = AllowAllStreamAcceptor::default();
```
russelltg, line 35: This is defintely one of the frustrating things about you merging your PRs together, is somtimes my comments go unaddressed...

No big deal, I know I could be faster about reviews and I don't have a problem saying it twice--anyways `accept_params` will always be Accept, so no need for this here

srt-protocol/src/protocol/sender/buffer.rs
```diff
@@ -472,7 +470,7 @@ mod test {
             recv_tsbpd_latency: TSBPD,
             cipher: None,
             stream_id: None,
-            bandwidth: LiveBandwidthMode::default(),
+            bandwidth: Default::default(),
```
russelltg, line 473: Seems like this change is a loss of visual information, but not huge deal

russelltg:
@robertream, here's an initial review. I haven't reviewd all the files yet, but I had limited time and figured I'd at least send back my thoughts. 

I love AccessResponse being an Input. Much cleaner than what I had.

Can we delete Multiplex yet?

There's some issue with large allocations in some of your newest tests

Rust users are used to the builder pattern, no reason to hide it from them by returning a builder form SrtSocket's new--can we just call that function `builder` to make it clear what's going on? I've seen that pattern in, for example, Quinn and I think it's slick. And We should call the builder `SrtSocketBuilder`, as it is a builder.

srt-protocol/src/listener/session.rs
robertream: > @robertream, here's an initial review. I haven't reviewd all the files yet, but I had limited time and figured I'd at least send back my thoughts.
> 
> I love AccessResponse being an Input. Much cleaner than what I had.
> 
> Can we delete Multiplex yet?

Deleting Multiplex is the end goal, at least shortly after this PR if not part of it. I haven't looked into if the effort is non-trivial. I'm less concerned about getting this PR perfect, we can clean it up as we iterate, we don't have many people using the library... yet. :)

> There's some issue with large allocations in some of your newest tests
> 
> Rust users are used to the builder pattern, no reason to hide it from them by returning a builder form SrtSocket's new--can we just call that function `builder` to make it clear what's going on? I've seen that pattern in, for example, Quinn and I think it's slick. And We should call the builder `SrtSocketBuilder`, as it is a builder.

I am fine with this, I have a bias against using programming design pattern names in types. Anything in my domain model that ends with "or", "er" or "ar", I am highly skeptical of, unless it's essential to the domain language (e.g. the SRT spec actually mentions Sender/Receiver Listener/Caller). Since Builder is a convention in Rust, I'll concede this point: "when in Rome...". I'm glad you approve of factory methods for builders as a member of the struct that is being built. I'm fine with naming the method "builder".

robertream:
> @robertream, here's an initial review. I haven't reviewd all the files yet, but I had limited time and figured I'd at least send back my thoughts.
> 
> I love AccessResponse being an Input. Much cleaner than what I had.
> 
> Can we delete Multiplex yet?

Deleting Multiplex is the end goal, at least shortly after this PR if not part of it. I haven't looked into if the effort is non-trivial. I'm less concerned about getting this PR perfect, we can clean it up as we iterate, we don't have many people using the library... yet. :)

> There's some issue with large allocations in some of your newest tests
> 
> Rust users are used to the builder pattern, no reason to hide it from them by returning a builder form SrtSocket's new--can we just call that function `builder` to make it clear what's going on? I've seen that pattern in, for example, Quinn and I think it's slick. And We should call the builder `SrtSocketBuilder`, as it is a builder.

I am fine with this, I have a bias against using programming design pattern names in types. Anything in my domain model that ends with "or", "er" or "ar", I am highly skeptical of, unless it's essential to the domain language (e.g. the SRT spec actually mentions Sender/Receiver Listener/Caller). Since Builder is a convention in Rust, I'll concede this point: "when in Rome...". I'm glad you approve of factory methods for builders as a member of the struct that is being built. I'm fine with naming the method "builder".

srt-protocol/src/listener/session.rs
robertream: yup

srt-protocol/src/listener/mod.rs
robertream: good point, I haven't used HashMap much yet

srt-protocol/src/listener/statistics.rs
robertream: Honestly, I was just throwing some things in there as a POC. I'm happy to remove them until it's better thought through.

srt-protocol/src/listener/statistics.rs
robertream: yeah... (sheepishly) connection? Again, the statistics are not the intent of the PR, these could be removed for now. Feel free to reorganize or rename any of them, I'm not putting much effort into it.

srt-protocol/src/options/validation.rs
robertream: So there is a design tensions here I was trying to explore.
1) I wanted the type system to enforce/encode validation for configuration
2) I wanted to use bare structs, i.e. all pub fields,  potentially easily serializable. So, no traditional builder pattern, which is  used often in Rust to solve this problem, yet at the cost of excessive boiler plate code in to tightly control mutable access to configuration data.
3)  I was unsure if a near completely flat options/configuration data structure that mirrored the reference implementation would be sufficient or if one with a little more conceptual organization/structure would be more desirable. I am exploring the later, and when I started to write validation for the options based off of the reference implementation, I ran into cases where I needed to do validation across multiple parts of the data in different structs, so I introduced a composite validation type. If we just went with a flat structure, this wouldn't be necessary. But if you like the shape of the "options" module's data, this is kind of essential to making it all work.

Also, I agree this needs more documentation, but I'm actually not sure yet if this structure is desirable/needed. I'm more curious for yours or other's feedback (@aegroto?).

srt-protocol/src/options/mod.rs
robertream: Ah, I believe I thought this was for the buffer used to receive a packet from the UdpSocket API, not the underlying OS UDP socket buffer... I agree the later is very important! :)

srt-protocol/src/options/encryption.rs
robertream: I put this in after noticing in the reference implementation's configuration documentation stated that an unspecified size has some meaning for  how encryption settings get negotiated, so I started to go down the path of implementing that, but I believe there are probably better ways to represent this in the configuration now, so I'll likely pull this out before we merge.

srt-protocol/src/protocol/pending_connection/hsv5.rs
robertream: > 

yeah, sorry, I thought I actually removed this.. hmmm. No worries on he speed of PRs, this is a labor of inspiration, no deadlines.

srt-protocol/src/protocol/sender/buffer.rs
robertream: I started using the pattern elsewhere, so I just did this for consistency. I'm mostly indifferent, with a slight bias now for Default::default(). Inference, in general hides all kinds of visual information, more isn't always better.... :)

srt-protocol/src/packet/mod.rs
robertream: Ok, well right now it is unclear which error value indicates that the underlying channel/socket has been closed. Since this struct is intended to abstract what could be a channel or socket, it needs to mash the semantics of the two together in a useful but clear way. Tough I'm not going to die on this hill, I don't wish to make the result type Result<Option<Packet>, RecievePacketError>, so my idea was to have RecievePacketError include end of stream semantics. This is actually a rough edge I want to get right before merge. Because it caused subtle deadlocking bug in the MulticastListener.

srt-protocol/src/packet/mod.rs
robertream: Yeah, probably, I just wanted any old implementation to make the compiler happy.

srt-protocol/src/options/validation.rs
aegroto: I'm more for keeping it simple and have a flat struct with public fields, although this would require for sure the configuration to be consumed by the socket's initialization. I'm not sure this will make the socket configurable dynamically even after the communication has started, have you thought about this scenario yet?

srt-protocol/src/options/mod.rs
aegroto: I confirm these may be very important, most of the times the bottlenecks of my project were given by saturated socket buffers.

srt-protocol/src/options/encryption.rs
aegroto: What about using an Option<KeySize>?

robertream: Unspecified is the default, and there is default key size if one is unspecified. Where there are semantics in the domain beyond Some/None, I prefer to encode it in the type, e.g. I can implement Default, and as_usize() and the semantics for a default size will be accessible anywhere. Domain modeling in a type system is often a fight against the urge/bias towards "primitive obsession". Option<T> is a primitive, though surely way better than C/C++ *T

srt-protocol/src/options/validation.rs
robertream: When I read "simple" I reflexively reinterpret it as "easy" and "familiar", and I respect this perspective, ergonomics and familiarity are important. Though ease is often confounded with simple. Breaking apart pieces into coherent chunks could be more simple relative to the domain problem. Sometimes we are better off adapt our sense of ease/familiarity to the simplicity essential to the domain problem/model. See: https://www.infoq.com/presentations/Simple-Made-Easy/ To be clear, I'm not arguing this is one of those times, but thank you for indulging me soap box :)

russelltg:
I'm sure there's still cleanup work to do, but wanted to get this merged to move the project forward. Thanks so much for the huge contribution!!

Issue #75, opened 2021-01-21
Connection freezes in high_bandwidth


russelltg:
Steps to reproduce:

Set the default buffer sizes to be larger (this should be an option that we expose, but for now it isn't):

```bash
sudo sysctl -w net.core.rmem_default=8388608
sudo sysctl -w net.core.wmem_default=8388608
```

Run the high_bandwidth test:

```bash
cargo test --release --test=high_bandwidth -- --nocapture
```

And notice after a few seconds it freezes. You may have to tweak the bandwidth by changing the `RATE_MBPS` constant to get it to repro (higher if it's working properly, lower if it isn't delivering the requested bandwidth).

This may or may not be reproducible outside of Linux, haven't tried.

robertream:
Is this resolved with the latest build? The latest changes might have fixed this too.

russelltg:
This may be fixed, but the high_bandwidth test still freezes up when ran for long enough at high enough bandwidth

robertream:
I agree, though ensure you are running it in release mode, at high speeds and too low of a latency setting, it's highly sensitive to thread timing. It might be interesting to see if this behavior could be reproduced in a "faster than real time" test.

russelltg:
Yep, that would be great to repro in a deterministic way. I am running in release, I put my repro steps in the original issue

russelltg:
(Blocked on #119 )

Issue #119, opened 2021-09-12
Implement Packet Pacing and Live Congestion Control (LiveCC) according to the latest spec.


robertream:
https://datatracker.ietf.org/doc/html/draft-sharabayko-srt-00#section-5.1

robertream:
Besides RTO, we should also make sure the other ARQ timers are consistent with the latest spec.

PR #136, opened 2021-11-24
add test to repro high bandwidth freeze easily

russelltg:
I've added a deterministic test to repro #75

### DIFF
--- a/srt-protocol/src/protocol/receiver/mod.rs
+++ b/srt-protocol/src/protocol/receiver/mod.rs
@@ -160,6 +160,7 @@ impl<'a> ReceiverContext<'a> {
                     BufferFull { .. } | PacketTooEarly { .. } | PacketTooLate { .. } => {
                         self.stats.rx_dropped_data += 1;
                         self.stats.rx_dropped_bytes += bytes;
+                        log::warn!("QoS warning--dropping data in the receive buffer");
                     }
                     DecryptionError(_) => {
                         self.stats.rx_decrypt_errors += 1;
--- a/srt-protocol/tests/lossy_deterministic.rs
+++ b/srt-protocol/tests/lossy_deterministic.rs
@@ -1,6 +1,7 @@
 // lossy tests based on protocol to be fully deterministic
 
 use std::cmp::min;
+use std::collections::VecDeque;
 use std::{
     str,
     time::{Duration, Instant},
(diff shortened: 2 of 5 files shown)


russelltg:
(I'm looking at it, so no need for you to dig into it @robertream. I'll lyk if I need help)

robertream:
> (I'm looking at it, so no need for you to dig into it @robertream. I'll lyk if I need help)

Nice, I love how side effect free state machine designs (or deterministic state transducer designs, to be more CS geeky) make stateful protocol code like this soooo easy to test. (state', input) -> (state'' output)

robertream:
> > (I'm looking at it, so no need for you to dig into it @robertream. I'll lyk if I need help)
> 
> Nice, I love how side effect free state machine designs (or deterministic state transducer designs, to be more CS geeky) make stateful protocol code like this soooo easy to test. (state', input) -> (state'' output)

BTW, I am working on making a state machine similar to DuplexConnection for the MultiplexListener too. So we can run simulations between callers and a listener in faster than real time as well. :)

russelltg:
Okay, figured out what's going on here--digging in the reference impl to see what they do:

The receive buffer size is not nearly large enough to deal with the latency*packet_rate product, so the receive buffer drops the packets, but the sender gets no notification of this--no NAK or ACK is sent for these packets. 

It seems like there are at least 3 issues going on here:
1. (the obvious one) buffer size is too small for this test
2. We don't retransmit on timeout (RTO in the spec)
3. We don't NAK packets that are dropped by the receive buffer--i'm not sure if this is proper of if you're just supposed to rely on RTO

robertream:
> Okay, figured out what's going on here--digging in the reference impl to see what they do:
> 
> The receive buffer size is not nearly large enough to deal with the latency*packet_rate product, so the receive buffer drops the packets, but the sender gets no notification of this--no NAK or ACK is sent for these packets.
> 
> It seems like there are at least 3 issues going on here:
> 
> 1. (the obvious one) buffer size is too small for this test
> 2. We don't retransmit on timeout (RTO in the spec)
> 3. We don't NAK packets that are dropped by the receive buffer--i'm not sure if this is proper of if you're just supposed to rely on RTO

2) is why I added #119

3) a NAK would trigger a resend, the spec says we need to send a “false” ACK, to move the forward

russelltg:
I think we do technically want a retransmit on full buffer--if the sender deems it's too late then so be it but retransmitting is fine.

I also assume we're supposed to use the reported buffer availability in the sender algorithm to make sure we don't overwhelm the buffer?

robertream:
> I think we do technically want a retransmit on full buffer--if the sender deems it's too late then so be it but retransmitting is fine. We should likely just look at the reference implementation

Ok, that’s reasonable

> I also assume we're supposed to use the reported buffer availability in the sender algorithm to make sure we don't overwhelm the buffer?

I assume that’s why the reference implementation was stalling when we didn’t send a valid buffer length, though I don’t recall seeing this in any of the specs.

russelltg:
Another interesting thing here--there is much more packet loss than I would expect--there's more packet loss than the connection even has, meaning we're not even beating normal UDP....yikes

robertream:
> Another interesting thing here--there is much more packet loss than I would expect--there's more packet loss than the connection even has, meaning we're not even beating normal UDP....yikes
> 
> See if you agree with me--the math is at the end of the new test

I got high_bandwidth to work consistently by lowering the SND period to the extreme, I think the throughput problems are all timing related, the stall is a protocol/buffer problem.

russelltg:
high_bandwidth (with increased buffer size, as in this branch) works at up to 90MB/s for me, which I think is pretty impressive

robertream:
> Another interesting thing here--there is much more packet loss than I would expect--there's more packet loss than the connection even has, meaning we're not even beating normal UDP....yikes
> 
> See if you agree with me--the math is at the end of the new test



> high_bandwidth (with increased buffer size, as in this branch) works at up to 90MB/s for me, which I think is pretty impressive

yeah, nice!

PR #138, opened 2021-11-29
introduce StreamId and PacketSocket

robertream:
- introduce StreamId and PacketSocket
- generalize the statistics WatchStream

These are a few changes worth keeping from the speculative listener redesign I've been kicking around.

Most notable is introducing PacketSocket, this shouldn't slow down execution at all and reduces unnecessary parametric polymorphism (i.e. impl trait fn arguments), I'm skeptical this will have a significant impact on compile time or binary size, but maybe it could.


srt-protocol/src/connection/mod.rs
```diff
@@ -303,6 +306,7 @@ impl DuplexConnection {
         self.timers.reset_exp(now);
 
         self.stats.rx_all_packets += 1;
+        self.stats.rx_all_bytes += size as u64;
```
russelltg, line 309: This is not correct--rx_all_bytes is supposed to include header sizes--I kinda prefer my approach to these metrics that I take in #134, but if you have a preference I'm willing to discuss.

srt-protocol/src/packet/error.rs
```diff
@@ -50,3 +51,37 @@ impl From<io::Error> for PacketParseError {
         PacketParseError::Io(s)
     }
 }
+
+impl Eq for PacketParseError {}
+
+impl PartialEq for PacketParseError {
```
russelltg, line 57: Would be good to have a comment here on how this deviates from a derive'd impl

srt-protocol/src/packet/stream_id.rs
```diff
@@ -0,0 +1,22 @@
+use std::{convert::TryFrom, string::FromUtf8Error};
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct StreamId(String);
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum StreamIdError {
```
russelltg, line 7: Ideally this would impl Error

srt-protocol/src/protocol/pending_connection/connect.rs
```diff
@@ -219,7 +223,7 @@ mod test {
             }),
         });
 
-        let resp = c.handle_packet((first, test_remote()), Instant::now());
+        let resp = c.handle_packet((0, first, test_remote()), Instant::now());
```
russelltg, line 226: These zeros are messy--another reason to go with `wire_size` in #134

russelltg (commented):
Overall looks pretty good--I like where this is going. I'm still unsure about the whole statistics interface but that's alright.

russelltg:
Oh and it looks like it panics on an invalid packet now

srt-protocol/src/connection/mod.rs
robertream: Ok, this was the bytes read in from receive buffer, I'm fine with pulling this out if you'd prefer.

srt-protocol/src/packet/error.rs
robertream: I couldn't derive it because io::Error does not implement Eq or PartialEq.

srt-protocol/src/packet/stream_id.rs
robertream: right

srt-protocol/src/protocol/pending_connection/connect.rs
robertream: yes, ok, I'll remove

robertream:
> Overall looks pretty good--I like where this is going. I'm still unsure about the whole statistics interface but that's alright.

Yeah, I'm not sold on the statistics interface either, I was messing with it here because I was exploring introducing multiplex listener statistics, so I made the watcher stream parametric, in order to easily reuse it.

srt-protocol/src/packet/error.rs
russelltg: I see. I actually feel iffy about making PacketParseError Eq+PartialEq considering io::Error isn't....digging thorugh it seems like the only thing that relies on that is the simulator's `ScheduledInput`, which I think it would be more correct anyways to just manually implement Eq+PartialEq on it to just consider the Instant. 

So my take is we should rip that out and just add a 

```rust
impl PartialEq for ScheduledInput {
    fn eq(&self, other: &Self) -> bool {
        self.0 == other.0
    }
}

impl Eq for ScheduledInput {}
```

russelltg: (imma go ahead and push that and merge this--we can talk if you don't agree)

PR #139, opened 2021-12-01
Listener

robertream:
@russelltg  This isn't done, but it's working so worth having you take a look.

There's a test in srt-tokio/src/listener/mod.rs that demonstrates the public API. I am mimicking the tokio::net::TcpListener API, yet this API allows for easy custom async enabled access control and session configuration.


russelltg:
Looks awesome, definitely something I was interested in having as well.

I'll take a look once you merge master into here

srt-protocol/src/listener/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    OpenConnectionFailed(SessionId),
+    DelegatePacketFailed(SessionId),
+}
+
+#[derive(Debug)]
+pub struct MultiplexListener {
+    settings: ConnInitSettings,
+    sessions: HashMap<SessionId, Option<Listen>>,
+    stats: ListenerStatistics,
+    stats_timer: Timer,
```
russelltg, line 75: "field is never read: `stats_timer`"

srt-protocol/src/listener/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    PendingConnectionMissing(SessionId),
+    ActiveConnectionMissing(SessionId),
+    OpenConnectionFailed(SessionId),
+    DelegatePacketFailed(SessionId),
+}
+
+#[derive(Debug)]
+pub struct MultiplexListener {
+    settings: ConnInitSettings,
+    sessions: HashMap<SessionId, Option<Listen>>,
```
russelltg, line 73: Would be good to have a comment here indicating a value of None means the connection is already established

srt-protocol/src/protocol/pending_connection/hsv5.rs
```diff
@@ -33,13 +32,36 @@ pub fn gen_hsv5_response(
         }
     };
 
+    let mut acceptor = AllowAllStreamAcceptor::default();
     let mut accept_params = match acceptor.accept(incoming.sid.as_deref(), from) {
```
russelltg, line 36: This can be simplified--AllowAllStreamAcceptor always returns Ok(AcceptParameters::default())

srt-tokio/src/multiplex.rs
```diff
@@ -1,9 +1,16 @@
-use std::{collections::HashMap, io, net::SocketAddr, time::Instant};
+use std::{
```
russelltg, line 1: We can probably remove multiplex now that we have SrtListener, which does the same thing, right?

srt-tokio/src/listener/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use super::net::PacketSocket;
+
+pub use srt_protocol::listener::ListenerStatistics;
+
+pub struct SrtListener {
+    settings: ConnInitSettings,
+    request_receiver: mpsc::Receiver<ConnectionRequest>,
+    statistics_receiver: watch::Receiver<ListenerStatistics>,
+    _task: JoinHandle<()>,
```
russelltg, line 22: What's the reason for holding onto this? Afaik JoinHandles don't have any special `Drop` behavior

srt-tokio/src/listener/request.rs
```diff
@@ -0,0 +1,95 @@
...
+use futures::{
+    channel::{mpsc, oneshot},
+    SinkExt,
+};
+use srt_protocol::settings::KeySettings;
+use srt_protocol::{connection::ConnectionSettings, listener::*, packet::*};
+
+use crate::{watch, SocketStatistics, SrtSocket};
+
+pub fn new_connection_request(
```
russelltg, line 13: Any reason why this isn't in `impl ConnectionRequest`?

srt-tokio/src/listener/request.rs
```diff
@@ -0,0 +1,95 @@
...
+        request,
+        settings,
+        output_data,
+        input_data,
+        statistics,
+    }
+}
+
+#[derive(Debug)]
+pub struct ConnectionRequest {
```
russelltg, line 33: I'm a bit unclear about what this struct does--does this only get constructed when there is an access controller installed? Or only when a valid StreamID is passed?

srt-protocol/src/protocol/pending_connection/listen.rs
```diff
@@ -182,42 +225,105 @@ impl Listen {
...
-            }
-            (ConclusionWait(state), ControlTypes::Handshake(shake)) => {
-                self.wait_for_conclusion(from, control.timestamp, state, shake, now, acceptor)
-            }
-            (InductionWait, control_type) | (ConclusionWait(_), control_type) => {
-                NotHandled(ConnectError::HandshakeExpected(control_type))
-            }
-        }
+        // TODO: handle StreamId parsing error
+        let sid = incoming.sid.clone().unwrap_or_default().into_bytes();
```
russelltg, line 238: Hmm, this seems to discard the difference between no StreamId an empty StreamId, which is a difference we do expose to the AccessController API (considering it takes an Option<&str>)

srt-protocol/src/protocol/pending_connection/listen.rs
```diff
@@ -182,42 +225,105 @@ impl Listen {
...
+        let stream_id = StreamId::try_from(sid).unwrap();
```
russelltg, line 239: This also seems to duplicate the work of validating that a StreamId is UTF-8--maybe StreamId should `impl From<String>`?

russelltg:
Alright, here's an initial review. Overall definitely the direction I want to go! Thanks so much for all the hard work.

srt-protocol/src/connection/mod.rs
```diff
@@ -23,13 +23,13 @@ use crate::{
     statistics::SocketStatistics,
 };
 
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
```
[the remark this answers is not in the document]
robertream: I don't disagree it might seem strange. I probably was forced to add this because this data structure ended up in a enum that derived Eq. But... in general I like the values in code (i.e. "Whole Values"/"Value Objects"/"Value Types") to all be equatable buy default. Settings and configuration are values too. This makes unit tests much more easy and ergonomic. If there was an alternative way to get easy deep structural comparisons in test, I'd be open to using that instead. The ability to automagically get pretty printing (Debug) and structural equality (Eq, PartialEq) are two big reasons I love Rust. The standard library has equatable collections, HashMaps, etc. mostly to enable this powerful feature. I've been spoiled before by using typed functional languages before, and got accustom to this style. Honestly, I'd prefer if Rust forced us to opt out of structural equality and pretty printing, for the occasional large stateful "object" with no legitimate approximation of value equality, or when a debug representation makes no sense. Even better though would be auto-derived structural diffing functionality, that's really what I want. Don't just test for equality, show me exactly what isn't the same. :)

srt-protocol/src/packet/mod.rs
```diff
@@ -138,4 +139,53 @@ impl Debug for Packet {
...
+        } else {
+            None
+        }
+    }
+}
+
+// NOTE: Eq, PartialEq are only here to accommodate structural comparison in tests without losing
+//  the ability to surface network errors for logging purposes, improve on this as needed, but don't
+//  remove.
+impl Eq for ReceivePacketError {}
```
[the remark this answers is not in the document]
robertream: assert_eq! automatically pretty prints out the two values, this is valuable in quickly debugging test failures, I will use matches! on occasion when constructing an exemplar value is impossible, tedious, or non-essential, I don't want to lose the general ergonomics of testing just because io::Error was not implemented correctly :)

srt-protocol/src/listener/mod.rs
robertream: I hadn't fully implemented statistics yet. I wanted to circle back and flush out the options spike, which I am about to push more changes to that PR, I think it's ready for prime time.

srt-protocol/src/listener/mod.rs
robertream: I was going to introduce an explicit enum to delineate the meaning instead. Leverage the type system to declare meaning/intent where possible.

srt-protocol/src/protocol/pending_connection/hsv5.rs
robertream: right, I thought I had gotten rid of all of these

srt-tokio/src/listener/mod.rs
robertream: I was pondering how to handle Listener shutdown. Should we join on all the open socket handles? Again, this is a rough draft.

srt-tokio/src/listener/request.rs
robertream: I wanted to accommodate the value being held by another thread/task, potentially in a data structure of some kind, returning impl would unnecessarily force said data structure to be parametric/generic.

srt-tokio/src/listener/request.rs
robertream: It always gets passed, similar in some way to an HttpRequest. Of course the response is limited to accept and reject though, which entirely informed by and sympathetic to the semantics of the protocol.

srt-tokio/src/multiplex.rs
robertream: Yes, that was the intent.

srt-protocol/src/protocol/pending_connection/listen.rs
robertream: An empty StreamId, which is essentially a String with a max length constraint, is legitimate, just like an empty String is legit. There's no benefit to adding a second layer of "nothing" semantics to the type. I would change the AccesssController API to just use &str, actually the intent was to have it use StreamId, just, again, this is a rough draft. I might even just simplify the AccessController API to just a Fn(AccessControlRequest) -> AccessControlResponse, and maybe Box<dyn> it to reduce monomorphisms.

robertream: An empty StreamId, which is essentially a String with a max length constraint, is legitimate, just like an empty String is legit. There's no benefit to adding a second layer of "nothing" semantics to the type. I would change the AccesssController API to just use &str, actually the intent was to have it use StreamId, just, again, this is a rough draft. I might even just simplify the AccessController API to just a Fn(AccessControlRequest) -> AccessControlResponse, and maybe Box<dyn> it to reduce monomorphisms.

srt-protocol/src/protocol/pending_connection/listen.rs
robertream: Yeah, perhaps. The goal was to, ergonomically, ensure all StreamIds instances are never in an invalid state (i.e. longer than 512 characters), I'm open to any improvements towards that goal.

srt-protocol/src/protocol/pending_connection/listen.rs
russelltg: But on the packet level, there is a differentiation between empty streamid and no streamid, and is also exposed on the sender side.

srt-protocol/src/packet/mod.rs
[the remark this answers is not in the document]
russelltg: Works for me

srt-tokio/src/listener/request.rs
russelltg: I didn't mean returning impl Trait, just an associated method (inside an impl struct {}) block

robertream: Oh, right, I'll revisit that, I don't recall why I did it this way

srt-protocol/src/protocol/pending_connection/listen.rs
robertream: Ok, you persuaded me, I"ll change all the APIs to reflect this.

srt-protocol/src/protocol/pending_connection/listen.rs
robertream: Ok, I actually implemented this in the separate "options" PR. I'll probably have to fix this when I merge/rebase.

Issue #12, opened 2018-06-06
Configurable maximim packet size and max flow size


russelltg:
Right now they're hardcoded.

robertream:
I actually wired this up in the big "options" merge, but we need to actually implement the connection negotiation algorithm for them.

PR #141, opened 2021-12-11
Implement RTO

russelltg:
#119

- start on rto i think
- initial take on RTO


russelltg:
@robertream The tests don't pass yet, but do you think this is a reasonable way to implement this?

srt-protocol/src/packet/control/mod.rs
```diff
@@ -173,10 +175,8 @@ pub struct HandshakeControlInfo {
 
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct AckStatistics {
-    /// Round trip time
-    pub rtt: TimeSpan,
-    /// RTT variance
-    pub rtt_variance: TimeSpan,
+    /// Round trip time+variance
+    pub rtt: Rtt,
```
robertream, line 179: I like this

robertream:
> @robertream The tests don't pass yet, but do you think this is a reasonable way to implement this?

In the last major redesign I did on DuplexConnecition, I felt like SenderCongestionControl was likely to disappear soon. The actual congestion control algorithm requires a coordination between separate domain concepts, the most obvious being sender/receiver. I can see how, at some point, it made sense to organize all the parts of the sender part of the algorithm in to one place, but this ultimately seemed requiring information sharing across or "object"/struct boundaries, which created very "chaty" interactions (read this, update that, etc). There is an essential complexity of the whole SRT congestion control algorithm which span multiple distinct conceptual pieces of the model (i.e. at least Sender/Receiver, then buffers, timers, etc.). Have one spot in the code that claims to implement even just the Sender part of this algorithm, felt cumbersome, and actually misleading too. Did you consider adding RTO to the Timers struct? This is where I was anticipating it would end up. We already do calculations with RTT there. I was hoping to migrate all the essential timers in the system to there. We can create individual structs for the more complicated timers (like RTO) in order to make them easier to understand and test in isolation.

What do you think?

russelltg:
I would really like to keep the congestion control as separate as I would like to be able to add a file congestion control in the future. I agree that It's not the most compelling use case for SRT, but there's a high probability I'll be able to get sponsorship in the near future to add it :)

russelltg:
As far as it being in Timers, that would be reasonable to me, but may actually be more "chatty" than the current implementation

robertream:
> I would really like to keep the congestion control as separate as I would like to be able to add a file congestion control in the future. I agree that It's not the most compelling use case for SRT, but there's a high probability I'll be able to get sponsorship in the near future to add it :)

Ok, well, even then, I would still advocate using say one or more enums for the design rather than a polymorphism. Again, the algorithm touches multiple parts of the code, so did encryption and I did this there, propagating configuration to separate "objects" that implement the whole algorithm in separate pieces.

robertream:
Also, consider my design views as $.02. The first order is getting it working correctly :)

robertream:
> As far as it being in Timers, that would be reasonable to me, but may actually be more "chatty" than the current implementation. I'll look into moving it

Also, note that the Sender RTT needs to be an exponential weighted moving average. Here's an off the shelf implementation in Rust I was going to lean on. May not be worth pulling in the library just for this one bit of code though.

https://github.com/greyblake/ta-rs/blob/master/src/indicators/exponential_moving_average.rs

russelltg:
Hmm but if rtt is straight from the receiver it's already filtered right? So this would be an exponential weighted average of an exponential weighted average?

robertream:
> Hmm but if rtt is straight from the receiver it's already filtered right? So this would be an exponential weighted average of an exponential weighted average?

Ha, good point, maybe I read it wrong.

russelltg:
@robertream This is ready for a proper review. I did ignore one test, it would be awesome if you could take a peek, I'm a bit lost on what the desired behavior is now.

robertream:
> > Hmm but if rtt is straight from the receiver it's already filtered right? So this would be an exponential weighted average of an exponential weighted average?
> 
> Ha, good point, maybe I read it wrong.



I didn't read it wrong. This is from the documentation:
https://datatracker.ietf.org/doc/html/draft-sharabayko-srt-00#section-5.1.2

```
   RTO is the amount of time within which an acknowledgement is expected
   after a data packet is sent out.  If there is no ACK after this
   amount of time has elapsed, a timeout event is triggered.  Since SRT
   only acknowledges every SYN time (Section 4.8.1), the value of
   retransmission timeout is defined as follows:

   RTO = RTT + 4 * RTTVar + 2 * SYN

   where RTT is the round-trip time estimate, in microseconds, and
   RTTVar is the variance of RTT estimate, in microseconds, reported by
   the receiver and smoothed at the sender side (see Section 3.2.4,
   Section 4.10).  Here and throughout the current section, smoothing
   means applying an exponentially weighted moving average (EWMA).
```

Perhaps the Receiver is actually supposed to send a non-smoothed estimate back? Though this is a draft, and I haven't looked into the reference implementation...

robertream:
@russelltg I just pushed a suggested refactor to your implementation. I also added a unit test to the send buffer covering RTO retransmission. Feel free to throw it out or riff of it.

russelltg:
Ok! Where did you push it?

robertream:
> Ok! Where did you push it?

Pushed it to my fork on accident.

robertream:
Ok, the build was working locally when I pushed last night, I'll try to sort this out.

robertream:
Ok, looks like I didn't run the tests after what seemed like an innocuous a clippy change.

It is potentially now doing a scan of the whole send buffer every time through the loop here in the general case. This needs to be optimized by caching the SeqNumber of the next RTO timeout in the send buffer, just wanted to get feedback before I invested further, regardless the tests I added and updated should all be legit either way.

russelltg:
Yeah, I think the heap was probably overkill.

My thought on the heap was that the RTO timer is complex, and packets can be retransmitted due to RTO more than once, so I wasn't sure how easy it was to simplify it. Figured correctness first performance later. However, of course linear search is correct :). That being said, I think just a variable tracking the next RTO is not correct, as packets can be retransmitted more than once. Hence why I reached for a heap.

Why is there no RTO on packets retransmitted via loss list?

I like storing the RTO in the buffer, I think it's a bit less pure but certainally is cleaner than having to pass that everywhere. 

Thank for fixing that test. 

What's your thoughts on TimeStamp vs Instant in implementation?

I was sticking with Instant because it's absolute and not relative and doesn't roll over. However, the nanosecond precision is overkill for us, but I'm sure computation overhead wasn't your concern?

russelltg:
I think a middleground for implementation of this would just be to track the very next RTO timeout and then just do a linear rescan when RTO occurs or when that timer times out. Could look like:

On packet send: next_rto = min(next_rto, my_rto)
When next_rto <= now: resend if timer is still valid (typically won't), and then rescan

That way it's only a single linear scan every RTO seconds which is quite infrequent

robertream:
> I think a middleground for implementation of this would just be to track the very next RTO timeout and then just do a linear rescan when RTO occurs or when that timer times out. Could look like:
> 
> On packet send: next_rto = min(next_rto, my_rto) When next_rto <= now: resend if timer is still valid (typically won't), and then rescan
> 
> That way it's only a single linear scan every RTO seconds which is quite infrequent

Yes, that could work, I also just optimized using a for real priority queue. What do you think?

robertream:
> Yeah, I think the heap was probably overkill.
I didn't think it was quite overkill, just had to hack on it a bit to come back around. What we want is a keyed priority queue, not just a heap.

> Why is there no RTO on packets retransmitted via loss list?
There should be, I was just mirroring your implementation. I fixed it in the last push.

> I like storing the RTO in the buffer, I think it's a bit less pure but certainally is cleaner than having to pass that everywhere.
You mean RTT, right?

> What's your thoughts on TimeStamp vs Instant in implementation?
>
> I was sticking with Instant because it's absolute and not relative and doesn't roll over. However, the nanosecond precision is overkill for us, but I'm sure computation overhead wasn't your concern?

Computation overhead wasn't my concern. The time calculations required for RTO will all fall will be within the modulo window and we have tests ensure that TimeStamp works correctly within the modulo window on rollover boundaries, so I'm also not concerned about rollover. I just didn't want the extra noise in the SendBuffer interface, requiring both ts_now and now in several places. I also didn't want to add the TimeBase back to the SendBuffer. We could revisit it later, but leaving it this way shouldn't noticeably affect behavior at all.

robertream:
@russelltg I just fixed the last of the high_bandwidth stalls with my last push, I would have created a separate PR, but the conflicts were non-trivial, I'm going to approve this PR so we can merge it, as long as you are happy with the implementation for now. We can always go back and change it later...

robertream (approved):

robertream:
@russelltg I've added an implementation of #12 to the PR now too...

russelltg:
The flow size calculation I think is not quite right-- https://github.com/haivision/srt/blob/8f169a9445a52f27f074cbc7b5f2bf92dd24e8e2/srtcore/socketconfig.h#L244

But i'm not totally sure what is right...

robertream:
> The flow size calculation I think is not quite right-- https://github.com/haivision/srt/blob/8f169a9445a52f27f074cbc7b5f2bf92dd24e8e2/srtcore/socketconfig.h#L244
> 
> But i'm not totally sure what is right... I'm not sure why it's involing the receive bfufer size but idk

iFlightFlagSize is actually SRTO_FC, a configuration option

That calculation is just putting a ceiling on the window, but it also appears to actually be dead code as well.

Here's what the options documentation has to say about SRTO_FC:
https://github.com/Haivision/srt/blob/224042a5867b11a06ddb5fb0f40d93ffb2b22494/docs/API/API-socket-options.md#SRTO_FC

It's a setting that affects the sender, but it is supposed to be limited by the peer's (meaning the other host's) receive buffer size, which makes some sense, though the receive buffers of a duplex connection could end up being configured differently, right? So this could be a hole in the logic for the session negotiation phase, where I don't think it enforces this limit correctly.

All this to say, I stand by my implementation and it also it appears to work as expected while testing it. Much better than the random complete stall we were seeing with high bandwidth if buffer limits were exceeded.

russelltg:
Ok. Merging

PR #147, opened 2021-12-13
Add ts_streamer example--doesn't work quite yet

russelltg:


Lighty0410:
Hey! Is it much of a hassle to ask for some code clarification regarding this PR right there in the review section? I have a very similar project but based on the draft I've done things differently. Thanks in advance !

russelltg:
Yeah, now would be an even better time now that it works :)

srt-tokio/examples/ts_streamer.rs
```diff
@@ -0,0 +1,159 @@
...
+    struct WriteBridge(Sender<(Instant, Bytes)>);
+    impl Write for WriteBridge {
+        fn write(&mut self, w: &[u8]) -> Result<usize, std::io::Error> {
+            for chunk in w.chunks(1316) {
+                // NOTE: Instant::now() is not ideal
+                // This should be directly derived from the PTS of the packet, which is not availabe here,
+                // it would require some refactoring
+                if self
+                    .0
+                    .try_send((Instant::now(), Bytes::copy_from_slice(chunk)))
```
Lighty0410, line 38: Oh, now i see. It's the exact same way i do in my own project. 
But previously you calculated a delta between current and the previous pts and then assign it to the latest instant. 
Can you elaborate on why it's recommended over the plane and simple `Instant::now` ?

russelltg: It's not a huge deal for this use case I believe, as the player obviously will make sure that the packets are displayed at the right time. But essentially how this is right now, there are two places for clock error--here and in the TSBPD code, versus if these are "correct" timestamps you'll get more consistent timing out of TSBPD as it won't include any error from here

Lighty0410: > It's not a huge deal for this use case I believe, as the player obviously will make sure that the packets are displayed at the right time. But essentially how this is right now, there are two places for clock error--here and in the TSBPD code, versus if these are "correct" timestamps you'll get more consistent timing out of TSBPD as it won't include any error from here

Got it. Thanks a lot for the explanation!

PR #153, opened 2021-12-24
limit send buffer size

### LINKED ISSUE #137: Limit send buffer size
Like the receiver buffer, we need to have a configurable limit of how many packets (or bytes?) to have in the send buffer before dropping them on the floor. 

Part of this is adding this to the tokio interface by adding backpressure (not letting someone add a packet to the buffer when its full)

robertream:
Closes #137

### DIFF
--- a/srt-protocol/src/settings/connection.rs
+++ b/srt-protocol/src/settings/connection.rs
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 use rand::random;
 
-use crate::options;
+use crate::{options, packet::Packet};
 
 use super::*;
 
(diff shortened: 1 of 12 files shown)


srt-protocol/src/settings/connection.rs
```diff
@@ -59,6 +61,7 @@ impl From<options::SocketOptions> for ConnInitSettings {
             bandwidth: options.sender.bandwidth,
             statistics_interval: options.session.statistics_interval,
             recv_buffer_size: options.receiver.buffer_size / options.session.max_segment_size,
+            send_buffer_size: options.sender.buffer_size / options.session.max_segment_size,
```
russelltg, line 64: Is buffer_size supposed to be how many data payload bytes? If so it should be `buffer_size / (mss - headers)`

russelltg:
overall looks great. just the one clarification

srt-protocol/src/settings/connection.rs
robertream: Good point, I checked the options documentation and the receive buffer size limit should also be calculated this way too.


> SRTO_RCVBUF
> 
> Receive Buffer Size, in bytes. Note, however, that the internal setting of this value is in the number of buffers, each one of size equal to SRT payload size, which is the value of SRTO_MSS decreased by UDP and SRT header sizes (28 and 16). The value set here will be effectively aligned to the multiple of payload size.

robertream:
@russelltg ok, I also fixed the max receive buffer size calculation too

PR #154, opened 2021-12-24
os bufsize

russelltg:
Also added the option for linger, which is not actually implemented in this PR

I wasn't exactly sure where to put the options, was that the logical place? The distinctions seem fuzzy to me

### DIFF
--- a/srt-protocol/src/options/connect.rs
+++ b/srt-protocol/src/options/connect.rs
@@ -80,3 +80,20 @@ impl Validation for Connect {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn ttl_validate() {
+        assert_eq!(
+            Connect {
+                ip_ttl: 0,
+                ..Default::default()
+            }
+            .is_valid(),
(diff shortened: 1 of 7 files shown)


srt-protocol/src/options/connect.rs
```diff
@@ -23,6 +23,35 @@ pub struct Connect {
     ///
     /// The default value is 0x010000 (SRT v1.0.0).
     pub min_version: SrtVersion,
+
+    /// SRTO_UDP_RCVBUF
+    ///
+    /// UDP Socket Receive Buffer Size. Configured in bytes, maintained in packets based on MSS value.
+    /// Receive buffer must not be greater than FC size.
```
robertream, line 30: You should add validation for this, and I'm guessing the send buffer ought to have the same limit enforced too.

robertream (approved):
Looks good to me.
The general rule I was applying to the options was Connect was just for settings for dealing with establishing a connection, Session was for general settings once the connection was established. Sender and Receiver were for SRT settings specific to sender and receiver.

I'm still thinking a flat data structure might be better ergonomically so I'm not too concerned about where these land right now.... :)

Issue #144, opened 2021-12-12
duplicate the srt-live-transmit srt url syntax


robertream:
https://github.com/Haivision/srt/blob/master/docs/apps/srt-live-transmit.md#medium-srt

russelltg:
We should probably keep this open until srt-live-transmit supports them

PR #156, opened 2021-12-27
SrtUri

robertream:
@russelltg

I've started on #144 though it is a work in progress still. There are a few more parameters left to implement, but I am pretty confident with the direction this is going. I haven't integrated the SrtUri with anything yet, so it wouldn't be a breaking change to merge this. This will be helpful in maintaining compatibility with the reference implementation CLI URI syntax. We can accommodate more than one key name per option, so it won't impede us from improving on the URI syntax while maintaining backwards compatibility.

### DIFF
--- a/srt-protocol/Cargo.toml
+++ b/srt-protocol/Cargo.toml
@@ -24,6 +24,7 @@ sha-1 = "0.10.0"
 streaming-stats = "0.2.3"
 take-until = "0.1.0"
 thiserror = "1.0.30"
+url = "=2.1.0" # https://github.com/servo/rust-url/issues/581
 
 [dependencies.aes]
 features = ["ctr"]
--- a/srt-protocol/src/options/address.rs
+++ b/srt-protocol/src/options/address.rs
@@ -1,37 +1,142 @@
 use std::{
-    convert::TryFrom,
-    net::{AddrParseError, SocketAddr},
-    net::{IpAddr, Ipv4Addr},
+    convert::{TryFrom, TryInto},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    num::ParseIntError,
+    str::FromStr,
 };
 
-use derive_more::*;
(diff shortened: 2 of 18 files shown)


robertream:
@russelltg influenced by the URL model, I decided to improve on SocketAddress to include domain host names, I am even more confident this should be merged now

russelltg:
Maybe I'm missing something but why do we want to include host names? ToSocketAddrs already can handle that, and the protocol doesn't care about hostnames (like TLS does).....what's the benefit?

robertream:
> Maybe I'm missing something but why do we want to include host names? ToSocketAddrs already can handle that, and the protocol doesn't care about hostnames (like TLS does).....what's the benefit?

The advantage is that the full BindOptions enum can be built up and validated without having to actually resolve domain names, thus keeping the protocol library separate from the I/O libraries, but keep common data types for configuration for all integrations (e.g. tokio/std::net/GStreamer etc.). I envisioned the SRT URI being reused for configuration in multiple contexts, not just srt-transmit, and I was attempting to use the URI for configuration in another context, which motivated this.

russelltg:
But the protocol library only cares about IP addresses, so isn't it both cleaner and more pure to keep domains away, and in IO land?

robertream:
> But the protocol library only cares about IP addresses, so isn't it both cleaner and more pure to keep domains away, and in IO land?

The protocol also doesn't know about UDP socket buffer sizes, but it's still in the options structs. I had initially attempted to keep that purity by excluding these i/o concepts from the options model, but you also found it useful to add them back in. I don't think this is much different. Perhaps there's a better way to solve the UDP socket configuration problem differently as well, but for now, I'm ok with the impurity/complexity. It made implementation easier this way.

russelltg:
I see, to natively support Uri's we need to have some sort of differentiation, I'm okay with it. I don't love it, and maybe longer term I'll find something I think is cleaner but for now it's fine.

Issue #53, opened 2020-06-28
C API


russelltg:
A C API made with cbindgen or something would make this crate more useful

PR #157, opened 2021-12-27
C API

russelltg:
#53


russelltg:
@robertream could you review the socket/listener changes in this PR?

The listener ones are the bigger ones--basically implementing the C api requires being able to close a listener _while_ polling for incoming connections, which I implemented by having the acceptor be taken out of the listener. I think overall a better API would be like the `TcpListener`, which has all the methods taking `&self`, so they can be executed concurrently. But a reasonable workaround for now?

robertream:
> @robertream could you review the socket/listener changes in this PR?
> 
> The listener ones are the bigger ones--basically implementing the C api requires being able to close a listener _while_ polling for incoming connections, which I implemented by having the acceptor be taken out of the listener. I think overall a better API would be like the `TcpListener`, which has all the methods taking `&self`, so they can be executed concurrently. But a reasonable workaround for now?

I don't like having all methods taking &self, at least for the tokio implementation, it'll require adding locks, if not a single lock around the whole data structure than a few around any mutable fields. The underlying channels can be easily accessed currently without locks, the Rust type system just gets in the way. This is one of those (still) unfortunately awkward aspects of lifetimes and references in Rust's type system: there's no way to have more than one simultaneous mutable reference on a struct without  breaking encapsulation and exposing all the fields, even if the struct is Pin'ed. A more idiomatic workaround might be to provide a split() method to explicitly break the structure apart into separate pieces when concurrency is desired. I know this is not consistent with the TcpListener API, but it would allow for concurrent access to be expressed in the Rust type system, without the need for an extra layer of locks (i.e. the underlying channels are already doing some locking of their own). Maybe make we could have an interface for the "read only" parts and the split() function can return an impl of the read only interface and an impl of the incoming stream? but I'm not sure, I'd have to play around with this a bit to see what looks/feels best. I don't like the incoming() function capturing the underlying channel, yet still allowing the function to be called again. Will it fail if it's called a second time? This seems awkward and strange, but this is just my first impression.

russelltg:
Yeah, I don't love it either--it was just the quickest way to get something that worked. The split idea seems reasonable, I'll give that a go

robertream:
Oh, I have another idea... what if the bind() method always returns both an SrtListener and an SrtIncoming instance? SrtListener just exposes close, settings, and statistics and other read only, concurrent friendly or clonable values, and SrtIncoming is just a wrapper for the underlying stream of ConnectionRequests? This is like with channel primitives, where they are always constructed split apart, because they are invariably used separately. I think this will be the most likely use case for SrtListener as well, we will almost always desire some way of shutting down the whole thing from a separate concurrent task/thread.

russelltg:
That's nice, okay will do

robertream:
> That's nice, okay will do

Boom! :)

Issue #112, opened 2021-06-26
Send shutdown on SIGINT for srt-transmit


russelltg:
Pretty self explainatory, currently it just waits for a timeout.

PR #165, opened 2022-01-28
transmit: Add SIGINT shutdown hook

Majored:
Hi all,

This PR is in relation to: https://github.com/russelltg/srt-rs/issues/112

A pretty simple addition; we periodically (every 500ms) check if the SIGINT flag has been set and break out of the piping loop if so (and thus the sinks are shutdown/closed).
Also slightly refactors the loop to be a bit more readable in the select! macro context.

### DIFF
--- a/srt-transmit/Cargo.toml
+++ b/srt-transmit/Cargo.toml
@@ -20,6 +20,7 @@ bytes = "1.0"
 anyhow = "1"
 pretty_env_logger = { version = "0.4", default-features = false }
 futures = { version = "0.3", default-features = false, features = ["std", "async-await"] }
+signal-hook = "0.3.12"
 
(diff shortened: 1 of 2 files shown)


russelltg:
Hello, welcome and why'd you close this immediately? Fyi: #131 has a start to this functionality, and has a unit test that may be functional--I don't remember exactly. My bad for not having my PR linked to that issue

Majored:
I hadn't seen https://github.com/russelltg/srt-rs/pull/131 originally, so closed this and was going to comment on it later to see what your intentions were (but didn't expect you to reply this quickly ;)).

Are you happy with the approach used? If so, I'll be happy to take a look at your unit test and incorporate it (along with looking into one for Windows given your comment).

PR #168, opened 2022-02-04
transmit: Add SIGINT shutdown hook (re)

Majored:
Hi all (again ;)),

This PR is in relation to https://github.com/russelltg/srt-rs/issues/112 and adds:
- A ZST error type representing a user interrupt error so that we can return a non 0/1 exit code.
- The actual SIGINT handling and termination with an exit code of 130.
- Unit tests for terminating an idle transmit (stdin, stdout) and where the transmit is doing actual work.

As was noted in https://github.com/russelltg/srt-rs/pull/131 's code comments, there doesn't seem to exist any crate for programmatically sending Ctrl+C events on Windows which I've added a comment about. There's a chance this could be handled via a raw winapi binding but I don't have access to a Windows machine currently to test. I'll look into it in the future if no one else has by then.

### DIFF
--- a/srt-transmit/Cargo.toml
+++ b/srt-transmit/Cargo.toml
@@ -20,6 +20,7 @@ bytes = "1.0"
 anyhow = "1"
 pretty_env_logger = { version = "0.4", default-features = false }
 futures = { version = "0.3", default-features = false, features = ["std", "async-await"] }
+signal-hook = "0.3.12"
 
 [dependencies.tokio]
 version = "1"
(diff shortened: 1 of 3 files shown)


Majored:
Looking at other actions, the Windows failure doesn't seem to be related to anything here.

As for why sigint_termination_idle() is failing, I'm unsure - I can't reproduce the failure on my end (Mac OSX M1). My only guess is that GitHub actions closes the child's stdin & stdout instantly somehow, thus the child exists with a 0 exit code rather than 130 (user interrupt). If so, I'm going to need to change the approach for that test...

PR #173, opened 2022-02-21
fix listener statistics

### LINKED ISSUE #172: Gathering statistics on SrtListener is blocked unless all clients are dropped
I'm trying to gather statistics from `SrtListener`. However, there's a problem. As soon as a client connects to the multiplex server, statistics gathering is blocked. And it's being locked until all clients are dropped. 
How to reproduce: 
1) Run the multiplexer with `RUST_LOG=info` - https://gist.github.com/Lighty0410/424dd2af7da8d45543ad208644b3b5ac 
2) Run the client (receiver) - https://github.com/russelltg/srt-rs/blob/main/srt-tokio/examples/receiver.rs 
3) Look at the multiplex server terminal. You can notice as soon as the client connects to the server there are no updates on the listener statistics. Statistics is gathering internally, however, the method call `binding.statistics().next()` is blocked.  
4) Shut down all clients. 
5) Look at the multiplex server terminal again. Now you can see the statistics get updates again. 

I tried selecting futures instead of spawning another tokio-thread, but the outcome is all the same.

robertream:
hopefully fixes #172

### DIFF
--- a/srt-tokio/src/listener/state.rs
+++ b/srt-tokio/src/listener/state.rs
@@ -4,9 +4,9 @@ use std::{
     time::{Duration, Instant},
 };
 
-use futures::{channel::mpsc, future::Fuse, prelude::*, select, SinkExt};
+use futures::{channel::mpsc, future::Fuse, prelude::*, select, SinkExt, FutureExt};
 use srt_protocol::{connection::Connection, listener::*, packet::*, settings::ConnInitSettings};
-use tokio::{sync::oneshot, time::sleep_until};
+use tokio::sync::oneshot;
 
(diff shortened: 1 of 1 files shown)


Lighty0410:
> hopefully fixes #172

Awesome! Now everything works. Thank you a lot! 
Is there any further testing required?

robertream:
> > hopefully fixes #172
> 
> Awesome! Now everything works. Thank you a lot! Is there any further testing required?

Maybe we could add something to a unit/integration test, but I don't think this is going to be something that changes often, so not sure it's super important now. It would have been better to have written the test first, then I wouldn't have pushed broken code... :)

Issue #183, opened 2023-03-01
Release new version on crates.io


BarthesSimpson:
Would you consider releasing an updated version on crates.io? The published version is quite old.

russelltg:
@MarcAntoine-Arnaud is actually the one who published it in the first place....It would be great if he could invite me to be another owner of the related crates? Or at least publish a new version....

nipierre:
Hi, I'm in direct contact with him so I'll speed things up.
Also, I will push a suggestion of crate push from tag in the CI, as we discussed some time ago.

MarcAntoine-Arnaud:
Hello,
I have invited you @russelltg.
But you've to manage versions in crates as v0.3.0 are already released.

Don't hesitate if you need some help.

BarthesSimpson:
Thanks for the fast response and action!

nipierre:
Now that the workflow is merged, can anyone test the release mechanism with a 0.4.0 ? :-)

russelltg:
After some minor fixes to the publishing logic, 0.4.1 (0.4.0 had a problem) is published!

PR #190, opened 2023-03-06
CI: publish automation

nipierre:
Addressing #183.

Should trigger the workflow on repo release and should add crate token as secret in the repo.


russelltg:
Is there a particular reason that `srt-c` is omitted from this, considering it has already been published as well?

nipierre:
My bad, I'll add it

nipierre:
@russelltg is it ok for you ? :-)

PR #197, opened 2023-03-23
add some failing tests

### LINKED ISSUE #195: Implement key size mismatch
`srt-rs` version: commit 1236c22
I have my listener setup like this:
```rs
let (_listener, mut receiver) = SrtListener::builder()
    .encryption(KeySize::AES256.as_raw(), "mypass")
    .bind(port)
    .await?;
while let Some(request) = receiver.incoming().next().await {
    let socket = match request.accept(None).await {
        Ok(socket) => socket,
        Err(err) => {
            error!("SRT connection error: {}", err);
            println!("{err:?}");
            continue;
        }
    };
    // ...
}
```
If I then use `ffplay` to connect like this:
```bash
ffplay "srt://127.0.0.1:51379/?passphrase=wrongpass"
```
I get this in my console output:
```
thread 'tokio-runtime-worker' panicked at 'not implemented: Key size mismatch', C:\Users\lucas\.cargo\git\checkouts\srt-rs-5cd6a57c40b8e1b5\1236c22\srt-protocol\src\protocol\pending_connection\hsv5.rs:72:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
SRT connection error: oneshot canceled
Custom { kind: NotConnected, error: Canceled }
```
If I recall correctly, the server should tell the receiver its key size. Is this a bug or has that not been implemented yet?
I know it says `not implemented`, but since "Encryption" has been marked as working in the README I thought I could use it.

Note:
Setting the key size manually in `ffplay` does work (default 0):
```bash
ffplay "srt://127.0.0.1:51379/?passphrase=mypass&pbkeylen=32"
```
But if you then use an invalid password:
```bash
ffplay "srt://127.0.0.1:51379/?passphrase=wrongpass&pbkeylen=32"
```
You'll get this error in your console output:
```
SRT connection error: oneshot canceled
Custom { kind: NotConnected, error: Canceled }
```

russelltg:
Fixes #195


srt-protocol/src/protocol/pending_connection/hsv5.rs
```diff
@@ -65,11 +67,12 @@ pub fn gen_access_control_response(
...
 
     // crypto
-    let cipher = match (&settings.key_settings, &incoming.ext_km) {
+    let cipher = match (&mut settings.key_settings, &incoming.ext_km) {
         // ok, both sides have crypto
         (Some(key_settings), Some(SrtControlPacket::KeyRefreshRequest(km))) => {
             if key_settings.key_size != incoming.key_size {
-                unimplemented!("Key size mismatch");
+                warn!("Key size mismatch: caller requested {:?}, listener was configured with {:?}. Selecting {:?}", incoming.key_size, key_settings.key_size, incoming.key_size);
+                key_settings.key_size = incoming.key_size;
```
Sculas, line 75: Is this intended behavior? This means the client can change the server's encryption settings to what it wants.
It can even set the key size to `0`, but thankfully that is then caught [here](https://github.com/russelltg/srt-rs/blob/c7a9e14257d12a3f7b82b926939d66e49a9365b9/srt-protocol/src/protocol/encryption/key.rs#L128) by defaulting to `Bytes16` instead of no encryption at all.

russelltg: Let me confirm what the ref impl does...

russelltg: Interesting, it seems to base it off SRTO_SENDER
https://github.com/haivision/srt/blob/39822840c506d72cef5a742d28f32ea28e144345/srtcore/core.cpp#L4927

russelltg: We don't currently implement SRTO_SENDER because it is only really needed for old SRT compatibility (new SRT is always full duplex). However the docs say that this is the only modern use of it https://github.com/Haivision/srt/blob/master/docs/API/API-socket-options.md#SRTO_SENDER

russelltg: For example:
```
$ srt-live-transmit "srt://:2222?pbkeylen=24&passphrase=password123" udp://127.0.0.1:3333 -a:no &
$ srt-live-transmit udp://:1111 "srt://127.0.0.1:2222?passphrase=password123&pbkeylen=16" -a:no
18:55:59.730216/SRT:RcvQ:w1!W:SRT.cn: processConnectResponse: PBKEYLEN conflict - keep 16; peer-advertised PBKEYLEN 24 rejected because Agent is SRTO_SENDER
```


```
$ srt-live-transmit udp://:1111 "srt://:2222?passphrase=password123&pbkeylen=16" -a:no &
$ srt-live-transmit "srt://127.0.0.1:2222?pbkeylen=24&passphrase=password123" udp://127.0.0.1:3333 -a:no
18:57:28.338138/SRT:RcvQ:w1!W:SRT.cn: processConnectResponse: PBKEYLEN conflict - OVERRIDDEN 24 by 16 from PEER (as AGENT is not SRTO_SENDER)

```

PR #200, opened 2023-04-01
Add additional TryFrom implementations for SocketAddress

cmleinz:
Adding the `TryFrom<Cow<...>>` and `TryFrom<String>` for `SocketAddress`

### DIFF
--- a/srt-protocol/src/options/address.rs
+++ b/srt-protocol/src/options/address.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Cow,
     convert::{TryFrom, TryInto},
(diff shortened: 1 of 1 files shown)


robertream (approved):
I feel like there could be a more generic implementation here that would avoid the need for specialized implementations, but until then, this seems reasonable.

cmleinz:
I agree it is not ideal, unfortunately there already exists a blanket implementation for TryFrom in core which incur [conflicts](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bf26834b9ecab5d847d0c477a820d14e) with generic implementations. 

Perhaps there's a way around there's a good way of deconflicting here but I'm not sure.

PR #206, opened 2023-06-22
Add AsyncRead and AsyncWrite Implementation

cmleinz:
Perhaps there's a reason for not doing this that I'm unaware of, but for my particular use-case it would be nice to have `SrtSocket` implement `AsyncRead` and `AsyncWrite`. Thanks for all the work on this library and let me know if there's another preferred method!

### DIFF
--- a/srt-tokio/src/socket/mod.rs
+++ b/srt-tokio/src/socket/mod.rs
@@ -14,7 +14,7 @@ use std::{
     time::Instant,
 };
 
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use futures::{
     channel::mpsc::{self, TrySendError},
     prelude::*,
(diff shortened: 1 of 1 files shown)


robertream:
Well this introduces buffer copying in order to conform to your desired interface. Since message buffers are essential to the current API design, the buffers need to be allocated on every write regardless, so a Reader/Writer interface never seemed necessary or desirable.

I'm ok merging this, but two things:
1) add test coverage for read and write.
2) add a warning/note in the documentation for the read/write implementations regarding the performance implications of buffer copying.

cmleinz:
This is a good point. Given that I don't know if it makes a lot of sense to move forward with this PR. The end user could always wrap the `SrtSocket` and implement these traits on their wrapped type.

PR #202, opened 2023-04-15
Fix: use more robust dns resolution

nipierre:
Closes #201.

Using [trust_dns_resolver crate](https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/index.html) and applying OS's configuration for DNS resolution.

It is resolving now:
`INFO  srt_media::source::srt_source > Opening source: srt://google.com:3001
 DEBUG srt_tokio::socket::call       > None:connect - SendPacket(({HS Induction from=SRT#74F897B9 UDT: Datagram ts=00:00.000000 dst=SRT#00000000}, 216.58.214.174:3001))`

Issue #201, opened 2023-04-15
tokio::net::lookup_host does not resolve 


nipierre:
Trying the new version of the crate, we encountered this error:

`Running target/debug/play-srt-source 'srt://my.domain.com:3001'
thread '<unnamed>' panicked at 'called Result::unwrap() on an Err value: Error { kind: InvalidInput, message: "invalid socket address" }', src/source_stream/srt_source_stream.rs:25:54`

This seems to be explained by the fact that `lookup_host` from `tokio` is pretty basic:
`This API is not intended to cover all DNS use cases. Anything beyond the basic use case should be done with a specialized library.`
(cf. https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html)

This should be replaced by a more capable library imho.

PR #207, opened 2023-07-27
srt-tokio: remove `trust-dns-resolver` dependency

Kijewski:
Without `trust-dns-resolver` a boatload of dependencies are removed from the project:

* `data-encoding`, `enum-as-inner`, `heck`, `hostname`, `ipconfig`, `ipnet`, `linked-hash-map`, `lock_api`, `lru-cache`, `match_cfg`, `matches`, `parking_lot`, `resolv-conf`, `scopeguard`, `trust-dns-proto`, `trust-dns-resolver`, `widestring`, `winreg`

If the user wants to use DNS-over-TLS, they can either lookup the name themself and supply a `SocketAddr` instead of a domain name, or they can set up their system resolver to use DNS-over-TLS.

This PR effectly reverts <https://github.com/russelltg/srt-rs/pull/202>, but fixes the bug in <https://github.com/russelltg/srt-rs/issues/201> instead. The problem was actually that Tokio's `lookup_host()` resolves to a `SocketAddr` instead of an `IpAddr`, so its argument needs a domain name and a port, not only a domain name.

### DIFF
--- a/srt-tokio/Cargo.toml
+++ b/srt-tokio/Cargo.toml
@@ -14,7 +14,6 @@ version = "0.4.1"
 bytes = "1"
 rand = "0.8"
 socket2 = "0.5"
-trust-dns-resolver = "0.22.0"
 
 [dependencies.ac-ffmpeg]
 optional = true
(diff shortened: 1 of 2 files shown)


Kijewski:
@nipierre, could you please verify if this change is fine or if it reintroduces the bug in #201 for you?

robertream (approved):

russelltg:
Thanks for the contribution!

nipierre:
> @nipierre, could you please verify if this change is fine or if it reintroduces the bug in #201 for you?

Fine by me, good job ! :)

Issue #166, opened 2022-01-31
Too late packets. 


Lighty0410:
Hey everyone! I'm extensively testing this library in real-world scenarios.  
And here's a weird flaw i found. Using srt-rs (sender) -> srt-rs (receiver) i get lots of too late packets whereas using srt-rs (sender) -> ffplay (receiver) i get none. 
How i tested: 
1) Run the multiplexor on the remote server. 
2) Run the srt-rs receiver locally. 
3) Run ffplay locally (tried with debugging too, no difference).
4) The srt-rs receiver throws lots of warnings (too late packets) whereas ffplay does not (**upd**, it does, read below).
The settings i use: 

The receiver: 
```
 ConnectionSettings { remote: remote_addr:remote_port, remote_sockid: SRT#SOME_SOCK, local_sockid: SRT#C91B0848, socket_start_time: Instant { tv_sec: 10475, tv_nsec: 806044913 }, rtt: 24.56258ms, init_seq_num: SeqNumber(763986792), max_packet_size: PacketSize(1316), max_flow_size: PacketCount(25600), send_tsbpd_latency: 120ms, recv_tsbpd_latency: 200ms, recv_buffer_size: PacketCount(8439), send_buffer_size: PacketCount(32), cipher: None, stream_id: None, bandwidth: Unlimited, statistics_interval: 1s }
```  

The sender: 
```
 ConnInitSettings { local_sockid: SRT#84B01C5B, key_settings: None, key_refresh: KeyMaterialRefreshSettings { period: 16777216, pre_announcement_period: 4096 }, send_latency: 200ms, recv_latency: 120ms, bandwidth: Unlimited, statistics_interval: 3ms, recv_buffer_size: PacketCount(8439), send_buffer_size: PacketCount(8439), max_packet_size: PacketSize(1316), max_flow_size: PacketCount(25600) }
```

This is the log i get on the receiver side: [log](https://gist.github.com/Lighty0410/357185716e8a817c6e5b370eac9e7ee9). 

As you can see the buffer is pretty huge and on top of that, i guess the recv/send latency of 200ms is more than enough for approx ~25ms rtt. The bandwidth of a stream is around 9.8 mbit/s. 
What's weird. Changing the latency/buffer doesn't seem to change much. All the same warnings with very tiny delay time (a couple of msec).
[...]

PR #208, opened 2023-09-11
Feat: drop too late packets option

nipierre:
Referred in #166.

Implemented too late packets drop disabling through socket options configuration.

### DIFF
--- a/srt-protocol/src/protocol/receiver/buffer.rs
+++ b/srt-protocol/src/protocol/receiver/buffer.rs
@@ -175,6 +175,7 @@ pub struct ReceiveBuffer {
     // first sequence number in the list
     seqno0: SeqNumber,
 
+    too_late_packet_drop: bool,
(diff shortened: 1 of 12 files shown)


srt-protocol/src/protocol/receiver/buffer.rs
```diff
@@ -262,9 +265,13 @@ impl ReceiveBuffer {
         let timestamp = match self.front_ts() {
             Some(timestamp) => timestamp,
             None => {
-                return match self.drop_too_late_packets(now) {
-                    Some(error) => Err(error),
-                    None => Ok(None),
+                if self.too_late_packet_drop {
+                    return match self.drop_too_late_packets(now) {
```
robertream, line 269: DRY this code: instead of repeating the `self.too_late_packet_drop` check, move it inside the `drop_too_late_packets` function

PR #214, opened 2024-04-15
fix: KeySize for as_usize was wrong, use call as_raw to fix it

MarcAntoine-Arnaud:


robertream:
how is this "wrong"? can you provide a test that demonstrates this ostensibly wrong behavior?

Issue #196, opened 2023-03-23
Handle server rejection properly


russelltg:
Repro:

```
cargo run -- udp://:1111 "srt://127.0.0.1:2222?passphrase=mypassasdaa&pbkeylen=16" &
cargo run -- "srt://:2222?passphrase=mypassasda&pbkeylen=16" -
```

gives

```
thread 'main' panicked at 'not yet implemented', /home/russell/srt-rs/srt-tokio/src/socket/listen.rs:37:29
```

PR #216, opened 2024-05-22
Expose the caller's specifed key size in ConnectionRequest

pierceforte:
Currently, it is not possible to retrieve the caller's specified key size when handling a `ConnectionRequest`.

This means we must defer to the library to handle the key size. This presents a number of issues:
1. The library cannot currently handle key size mismatches (https://github.com/russelltg/srt-rs/issues/195).
    - This is especially problematic because it means one caller's bad request can shut down the server and drop all other callers' connections.
2. The library does not currently handle rejection properly (https://github.com/russelltg/srt-rs/issues/196), which involves rejecting on the basis of the passphrase and its corresponding key size.
3. The server implementation cannot retrieve the key size despite it being advertised in the caller's request. This prevents any custom handling of the key size.

---

This PR adds a key size field to the `ConnectionRequest` type, which serves as a workaround for 1 and 2 and a solution for 3.

It also includes:
- a simple test
- an example multiplex server that enforces password authorization and prevents key size mismatch errors

### DIFF
--- a/srt-protocol/src/protocol/pending_connection/listen.rs
+++ b/srt-protocol/src/protocol/pending_connection/listen.rs
@@ -611,6 +611,6 @@ mod test {
             panic!("expected a ConnectionResult::RequestAccess");
         };
 
-        assert_eq!(request_access.key_size, send_key_size);
+        assert_eq!(request_access.key_size, hs_key_size);
     }
 }
--- a/srt-protocol/src/protocol/pending_connection/listen.rs
+++ b/srt-protocol/src/protocol/pending_connection/listen.rs
@@ -581,11 +581,11 @@ mod test {
             Ok((build_hs_pack(test_induction()), conn_addr())),
         );
 
-        let send_key_size = KeySize::AES256;
+        let hs_key_size = KeySize::AES256;
 
         let shake = HandshakeControlInfo {
             info: HandshakeVsInfo::V5(HsV5Info {
-                key_size: send_key_size,
+                key_size: hs_key_size,
                 ext_hs: Some(SrtControlPacket::HandshakeRequest(SrtHandshake {
                     version: SrtVersion::CURRENT,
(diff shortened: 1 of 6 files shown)


robertream:
Thank you! This is a great improvement. I can help you get this merged too. Two things though:
1) the build is broken, though it's not your fault. I have a PR that hopefully fixes this: #217
2) I appreciate the integration test coverage, and for this behavior a unit test in the protocol library is preferable for two reasons:
>  a) simple errors in logic are often better covered by fast isolation tests vs. much slower integration tests. Faster feedback cycles during development are desirable
>  b) testing at the protocol level provides coverage for all possible uses of the protocol (e.g. tokio vs c API, etc.)

Once the build is fixed, I'll get your changes merged. If you'd add a unit test as well, I'd appreciate that, but I won't block merging your PE on this.

robertream:
Ok, the macos build is failing, because of c library compilation failures, but we can deal with that later.

robertream:
You can quickly and easily run both clippy and fmt locally to avoid the CI build failures

pierceforte:
@robertream Hi, thank you for your help. I'll add the unit test later today – we can wait to merge until that's in.

Oddly, I'm not able to see these clippy warnings or correct the formatting with the cargo commands.

The clippy command seems to be getting stuck on the macos issues.

robertream:
These integration test failures we are seeing in CI here are the precise reason why I prefer determinist isolation tests ("unit tests") to non-deterministic integraion tests for coverage of determinist logic.

pierceforte:
Hi @robertream I updated the PR to remove the integration test and instead include a unit test. Please let me know if it looks ok.

pierceforte:
Hi @robertream, can you please cut a release once the macos build is fixed? Also, is there anything I can do to help with the fix?

